PR #10, opened 2022-07-05
Added possibility to use properties to configure connector

berndruecker:


nikku:
v0.2.0 of the SDK ships with [automatic configuration](https://github.com/camunda/connector-sdk/tree/main/runtime-job-worker#automatic-connector-discovery). Let us validate with users if additional investments such as property-based configuration are still necessary.

Closing this PR for now.

PR #59, opened 2022-08-05
fix: call Zeebe with result variable map

### LINKED ISSUE #45: Object returned by ConnectorFunction(..) must be 'root level object'
__Describe the Bug__

When I played around with ConnectorFunction in the first place I simply returned a String but this failed in ConnectorJobHandler with:

`    public Object execute(ConnectorContext context) throws Exception {
        String variables = context.getVariables();
        System.out.println("Variables as Json: " + variables);
        return "Bla";
    }`

`SEVERE: Failed to process job 2251799813686878
io.camunda.zeebe.client.api.command.ClientStatusException: Property 'variables' is invalid: Expected document to be a root level object, but was 'STRING'
        at io.camunda.zeebe.client.impl.ZeebeClientFutureImpl.transformExecutionException(ZeebeClientFutureImpl.java:93)
        at io.camunda.zeebe.client.impl.ZeebeClientFutureImpl.join(ZeebeClientFutureImpl.java:50)
        at io.camunda.connector.runtime.jobworker.ConnectorJobHandler.handle(ConnectorJobHandler.java:32)
        at io.camunda.zeebe.client.impl.worker.JobRunnableFactory.executeJob(JobRunnableFactory.java:44)`

__Expected Behavior__

The API should force returning an expected Object structure (String or other wrapped primitive data types make no sense at all but are allowed by the contract), e.g. by using some kind of ConnectorResult interface.

`    public ConnectorResult execute(ConnectorContext context) throws Exception {
        String variables = context.getVariables();
        System.out.println("Variables as Json: " + variables);
        return new MyResult("blubb");
    }`

__Environment__

- Library version: [0.1.0-SNAPSHOT]
- Camunda version: [e.g. 8.0.4 (self-managed)]

tmetzke:
## Description

Wraps the connector function result in a map before passing it
on to Zeebe. The result is only returned if a `resultVariable`
header is defined on the task (consistent with C8 SaaS Bridge behavior).

## Related issues

* related to #8 (has a broader scope including evaluating `resultExpression` with FEEL, for example)
* closes #45

### DIFF
--- a/runtime-job-worker/pom.xml
+++ b/runtime-job-worker/pom.xml
@@ -43,6 +43,12 @@
     </dependency>
 
     <dependency>
+      <groupId>org.assertj</groupId>
+      <artifactId>assertj-core</artifactId>
+      <scope>test</scope>
+    </dependency>
+
+    <dependency>
       <groupId>org.mockito</groupId>
       <artifactId>mockito-core</artifactId>
       <scope>test</scope>
--- a/runtime-job-worker/src/main/java/io/camunda/connector/runtime/jobworker/ConnectorConfig.java
+++ b/runtime-job-worker/src/main/java/io/camunda/connector/runtime/jobworker/ConnectorConfig.java
@@ -14,6 +14,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
 package io.camunda.connector.runtime.jobworker;
 
 import java.util.ArrayList;
@@ -27,10 +28,10 @@ public class ConnectorConfig {
   public static final Pattern ZEEBE_CONNECTOR_PATTERN =
       Pattern.compile("^ZEEBE_CONNECTOR_([^_]+)_TYPE$");
 
-  private String name;
-  private String type;
+  private final String name;
+  private final String type;
   private final String[] variables;
-  private String function;
+  private final String function;
 
   /**
    * Create a connector configuration.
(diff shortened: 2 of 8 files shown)


nikku:
@Oleksiivanov looking into your changes I could not really see how these simplified the tests.

Because of that I went ahead and built what I previously had envisioned (https://github.com/camunda/connector-sdk/pull/59#pullrequestreview-1066297308).

---

With the last two commits you'll see the following changes:

* Clearly structure the tests around different aspects (Secrets / Output)
* Get rid of endless mocking in favor of a minimal test utility

image

---

Could you @Oleksiivanov and @igpetrov have a look and see if these changes make sense to you? Again, my goal was to keep the tests maintainable + enable quick creation of further tests in the future.

nikku (approved):
Looks good now with https://github.com/camunda/connector-sdk/pull/59#issuecomment-1216465934.

nikku:
As per our [definition of done](https://github.com/camunda/connectors-internal-docs/blob/main/development/DEFINITION_OF_DONE.md#definition-of-done) this shall be cleaned up to separate the many concerns mixed in this PR:

* Migrate to static use of Mockito + other style upgrades
* Adding the test helpers
* Actually implementing + testing the feature

I'll spend a few minutes to see if I can do further cleanups. Afterwards I'll merge it.

PR #66, opened 2022-08-23
chore: implement result expression support

### LINKED ISSUE #8: Implement output mapping in Job Worker run-time
__Describe the problem__

We should implement all output mapping features supported by connectors:

image

Right now output mappings are not sufficently supported by the job worker run-time.

__Describe the solution__

* [x] Support `resultVariable` mapping (https://github.com/camunda/connector-sdk/pull/59)
* [ ] Support dynamic mapping
    * Extract using FEEL
    * Fail if output mapping fails and raise incident
    * Related https://github.com/camunda/connector-sdk/pull/66

igpetrov:
## Description

Implement result expression support

closes #8

### DIFF
--- a/runtime-job-worker/src/main/java/io/camunda/connector/runtime/jobworker/ConnectorJobHandler.java
+++ b/runtime-job-worker/src/main/java/io/camunda/connector/runtime/jobworker/ConnectorJobHandler.java
@@ -17,14 +17,19 @@
 
 package io.camunda.connector.runtime.jobworker;
 
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
 import io.camunda.connector.api.ConnectorContext;
 import io.camunda.connector.api.ConnectorFunction;
 import io.camunda.connector.api.ConnectorInput;
 import io.camunda.connector.api.SecretProvider;
 import io.camunda.connector.api.SecretStore;
+import io.camunda.connector.runtime.jobworker.feel.FeelEngineWrapper;
+import io.camunda.connector.runtime.jobworker.feel.FeelEngineWrapperException;
 import io.camunda.zeebe.client.api.response.ActivatedJob;
 import io.camunda.zeebe.client.api.worker.JobClient;
 import io.camunda.zeebe.client.api.worker.JobHandler;
+import java.util.HashMap;
 import java.util.Map;
 import java.util.Optional;
 import java.util.ServiceLoader;
@@ -36,9 +41,14 @@ public class ConnectorJobHandler implements JobHandler {
 
   private static final Logger LOGGER = LoggerFactory.getLogger(ConnectorJobHandler.class);
 
-  protected static final String RESULT_VARIABLE_HEADER_NAME = "resultVariable";
+  private static final String ERROR_CANNOT_PARSE_VARIABLES = "Cannot parse variables";
 
+  protected static final String RESULT_VARIABLE_HEADER_NAME = "resultVariable";
+  protected static final String RESULT_EXPRESSION_HEADER_NAME = "resultExpression";
+
+  private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
   private final ConnectorFunction call;
+  private final FeelEngineWrapper feelEngineWrapper;
 
   /**
    * Create a handler wrapper for the specified connector function.
@@ -47,6 +57,7 @@ public class ConnectorJobHandler implements JobHandler {
    */
   public ConnectorJobHandler(final ConnectorFunction call) {
     this.call = call;
+    this.feelEngineWrapper = new FeelEngineWrapper();
   }
 
   @Override
@@ -74,10 +85,29 @@ public class ConnectorJobHandler implements JobHandler {
 
   protected Map<String, Object> createOutputVariables(
       final Object responseContent, final Map<String, String> jobHeaders) {
-    return Optional.ofNullable(jobHeaders)
-        .map(headers -> headers.get(RESULT_VARIABLE_HEADER_NAME))
-        .map(resultVariableName -> Map.of(resultVariableName, responseContent))
-        .orElseGet(Map::of);
+    final Map<String, Object> outputVariables = new HashMap<>();
+    final var resultVariableName = jobHeaders.get(RESULT_VARIABLE_HEADER_NAME);
+    final var resultExpression = jobHeaders.get(RESULT_EXPRESSION_HEADER_NAME);
+
+    if (resultVariableName != null) {
+      outputVariables.put(resultVariableName, responseContent);
+    }
+
+    Optional.ofNullable(resultExpression)
+        .map(expression -> feelEngineWrapper.evaluateToJson(expression, responseContent))
+        .map(json -> parseJsonVarsAsMapOrThrow(json, resultExpression))
+        .ifPresent(map -> outputVariables.putAll(map));
+
+    return outputVariables;
+  }
+
+  private Map<String, Object> parseJsonVarsAsMapOrThrow(
+      final String jsonVars, final String expression) {
+    try {
+      return OBJECT_MAPPER.readValue(jsonVars, Map.class);
+    } catch (JsonProcessingException e) {
+      throw new FeelEngineWrapperException(ERROR_CANNOT_PARSE_VARIABLES, expression, jsonVars, e);
+    }
   }
 
   protected SecretProvider getSecretProvider() {
--- a/runtime-job-worker/src/main/java/io/camunda/connector/runtime/jobworker/feel/FeelEngineWrapper.java
+++ b/runtime-job-worker/src/main/java/io/camunda/connector/runtime/jobworker/feel/FeelEngineWrapper.java
@@ -1,0 +1,96 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH
+ * under one or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information regarding copyright
+ * ownership. Camunda licenses this file to you under the Apache License,
+ * Version 2.0; you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.camunda.connector.runtime.jobworker.feel;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.module.scala.DefaultScalaModule$;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Objects;
+import org.camunda.feel.FeelEngine;
+import org.camunda.feel.impl.SpiServiceLoader;
+import scala.jdk.javaapi.CollectionConverters;
+
+public class FeelEngineWrapper {
+
+  static final String RESPONSE_MAP_KEY = "response";
+  static final String ERROR_VARIABLES_MUST_NOT_BE_NULL = "variables cannot be null";
+  static final String ERROR_EXPRESSION_EVALUATION_FAILED = "expression evaluation failed";
+
+  private final FeelEngine feelEngine;
+  private final ObjectMapper objectMapper;
+
+  public FeelEngineWrapper() {
+    this(
+        new FeelEngine.Builder()
+            .valueMapper(SpiServiceLoader.loadValueMapper())
+            .functionProvider(SpiServiceLoader.loadFunctionProvider())
+            .build(),
+        new ObjectMapper().registerModule(DefaultScalaModule$.MODULE$));
+  }
+
+  public FeelEngineWrapper(final FeelEngine feelEngine, final ObjectMapper objectMapper) {
+    this.feelEngine = feelEngine;
+    this.objectMapper = objectMapper;
+  }
+
+  private static String trimExpression(final String expression) {
+    var feelExpression = expression.trim();
+    if (feelExpression.startsWith("=")) {
+      feelExpression = feelExpression.substring(1);
+    }
+    return feelExpression.trim();
+  }
+
+  private static scala.collection.immutable.Map<String, Object> toScalaMap(
+      final Map<String, Object> responseMap) {
+    final HashMap<String, Object> context = new HashMap<>(responseMap);
+    context.put(RESPONSE_MAP_KEY, responseMap);
+    return scala.collection.immutable.Map.from(CollectionConverters.asScala(context));
+  }
+
+  @SuppressWarnings("unchecked")
+  private static Map<String, Object> ensureVariablesMap(final Object variables) {
+    return (Map<String, Object>)
+        Objects.requireNonNull(variables, ERROR_VARIABLES_MUST_NOT_BE_NULL);
+  }
+
+  public String evaluateToJson(final String expression, final Object variables) {
+    try {
+      var variablesAsMap = ensureVariablesMap(variables);
+      var variablesAsMapAsScalaMap = toScalaMap(variablesAsMap);
+      var result = feelEngine.evalExpression(trimExpression(expression), variablesAsMapAsScalaMap);
+      if (result.isRight()) {
+        return resultToJson(result.right().get());
+      } else {
+        throw new RuntimeException(result.left().get().message());
+      }
+    } catch (Exception e) {
+      throw new FeelEngineWrapperException(
+          ERROR_EXPRESSION_EVALUATION_FAILED, expression, variables, e);
+    }
+  }
+
+  private String resultToJson(final Object result) {
+    try {
+      return objectMapper.writeValueAsString(result);
+    } catch (final JsonProcessingException e) {
+      throw new RuntimeException(e);
+    }
+  }
+}
--- a/runtime-job-worker/src/test/java/io/camunda/connector/runtime/jobworker/ConnectorJobHandlerTest.java
+++ b/runtime-job-worker/src/test/java/io/camunda/connector/runtime/jobworker/ConnectorJobHandlerTest.java
@@ -17,10 +17,14 @@
 
 package io.camunda.connector.runtime.jobworker;
 
+import static io.camunda.connector.runtime.jobworker.ConnectorJobHandler.RESULT_EXPRESSION_HEADER_NAME;
+import static io.camunda.connector.runtime.jobworker.ConnectorJobHandler.RESULT_VARIABLE_HEADER_NAME;
 import static org.assertj.core.api.Assertions.assertThat;
 
 import io.camunda.connector.api.ConnectorFunction;
 import io.camunda.connector.api.SecretProvider;
+import java.util.Collections;
+import java.util.HashMap;
 import java.util.Map;
 import org.junit.jupiter.api.Nested;
 import org.junit.jupiter.api.Test;
@@ -31,18 +35,17 @@ public class ConnectorJobHandlerTest {
   class Secrets {
 
     @Test
-    public void shouldReplaceSecretsViaSpiLoadedProvider() throws Exception {
+    public void shouldReplaceSecretsViaSpiLoadedProvider() {
       // given
       var jobHandler =
           new ConnectorJobHandler(
-              (context) -> {
-                return context
-                    .getSecretStore()
-                    .replaceSecret("secrets." + TestSecretProvider.SECRET_NAME);
-              });
+              (context) ->
+                  context
+                      .getSecretStore()
+                      .replaceSecret("secrets." + TestSecretProvider.SECRET_NAME));
 
       // when
-      var result = JobBuilder.create().withResultHeader("result").execute(jobHandler);
+      var result = JobBuilder.create().withResultVariableHeader("result").execute(jobHandler);
 
       // then
       assertThat(result.getVariable("result")).isEqualTo(TestSecretProvider.SECRET_VALUE);
@@ -53,14 +56,13 @@ public class ConnectorJobHandlerTest {
       // given
       var jobHandler =
           new TestConnectorJobHandler(
-              (context) -> {
-                return context
-                    .getSecretStore()
-                    .replaceSecret("secrets." + TestSecretProvider.SECRET_NAME);
-              });
+              (context) ->
+                  context
+                      .getSecretStore()
+                      .replaceSecret("secrets." + TestSecretProvider.SECRET_NAME));
 
       // when
-      var result = JobBuilder.create().withResultHeader("result").execute(jobHandler);
+      var result = JobBuilder.create().withResultVariableHeader("result").execute(jobHandler);
 
       // then
       assertThat(result.getVariable("result")).isEqualTo("baz");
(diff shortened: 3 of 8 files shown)


runtime-job-worker/src/test/java/io/camunda/connector/runtime/jobworker/FeelEngineWrapperTest.java
```diff
@@ -0,0 +1,157 @@
...
+    final String evaluatedResultAsJson =
+        objectUnderTest.evaluateFeelExpressionToJson(resultExpression, variables);
+
+    // then
+    Map<String, Object> jsonVars = UTILITY_JSON_MAPPER.readValue(evaluatedResultAsJson, Map.class);
+    assertThat(jsonVars.size()).isEqualTo(1);
+    assertThat(jsonVars).containsKey("processedOutput");
+    assertThat(jsonVars.get("processedOutput")).isInstanceOf(Map.class);
+    Map resultEntry = (Map) jsonVars.get("processedOutput");
+    assertThat(resultEntry).containsEntry("statusCode", "200 OK");
```
tmetzke, line 58: How about we simplify those tests and compare the result against the expected JSON String?
```suggestion
    assertThat(evaluatedResultAsJson).isEqualTo("{\"processedOutput\":{\"statusCode\":\"200 OK\"}");
```

runtime-job-worker/src/test/java/io/camunda/connector/runtime/jobworker/ConnectorJobHandlerTest.java
```diff
@@ -88,23 +83,79 @@ public void shouldNotSetWithoutResultVariable() throws Exception {
...
+    var jobHandler =
+        new ConnectorJobHandler((context) -> Map.of("callStatus", Map.of("statusCode", "200 OK")));
+
+    // when
+    var result = JobBuilder.create().execute(jobHandler);
+
+    // then
+    assertThat(result.getVariables()).isEmpty();
+  }
+
```
tmetzke, line 130: I believe we can remove this as this is covered by `shouldNotSetWithoutResultVariable` (which we can probably rename to something like `shouldNotSetWithoutResultVariableAndExpression`.

```suggestion
```

tmetzke (dismissed):
Looks really good already! 👍 
I added some minor suggestions, mostly code style or other smaller suggestions.

runtime-job-worker/src/main/java/io/camunda/connector/runtime/jobworker/ConnectorJobHandler.java
```diff
@@ -74,10 +81,27 @@ public void handle(final JobClient client, final ActivatedJob job) {
...
-    return Optional.ofNullable(jobHeaders)
-        .map(headers -> headers.get(RESULT_VARIABLE_HEADER_NAME))
-        .map(resultVariableName -> Map.of(resultVariableName, responseContent))
-        .orElseGet(Map::of);
+    final Map<String, Object> outputVariables = new HashMap<>();
+    final var resultVariableName = jobHeaders.get(RESULT_VARIABLE_HEADER_NAME);
+    final var resultExpression = jobHeaders.get(RESULT_EXPRESSION_HEADER_NAME);
+
+    Optional.ofNullable(resultVariableName)
+        .ifPresent(variableName -> outputVariables.putAll(Map.of(variableName, responseContent)));
```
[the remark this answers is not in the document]
igpetrov: No objections.

runtime-job-worker/src/main/java/io/camunda/connector/runtime/jobworker/ConnectorJobHandler.java
```diff
@@ -74,10 +81,27 @@ public void handle(final JobClient client, final ActivatedJob job) {
...
+
+    Optional.ofNullable(resultExpression)
+        .map(x -> feelEngineWrapper.evaluateFeelExpressionToJson(x, responseContent))
+        .map(y -> parseJsonVarsAsMapOrThrow(y))
+        .ifPresent(z -> outputVariables.putAll(z));
```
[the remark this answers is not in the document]
igpetrov: Yep, somehow I missed that myself.

runtime-job-worker/src/main/java/io/camunda/connector/runtime/jobworker/ConnectorJobHandler.java
```diff
@@ -74,10 +81,27 @@ public void handle(final JobClient client, final ActivatedJob job) {
...
+
+    return outputVariables;
+  }
+
+  private Map<String, Object> parseJsonVarsAsMapOrThrow(final String jsonVars) {
+    try {
+      return OBJECT_MAPPER.readValue(jsonVars, Map.class);
+    } catch (JsonProcessingException e) {
+      throw new RuntimeException(e);
```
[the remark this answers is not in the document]
igpetrov: Fair point

runtime-job-worker/src/test/java/io/camunda/connector/runtime/jobworker/FeelEngineWrapperTest.java
igpetrov: That's not a bad idea, but, IMO, that causes too much pain when we do refactoring of respective code. What I see people doing in that case - once refactoring complete, people run tests and once it doesn't pass, they blindly copy-paste **expected** value.

So I would prefer to keep as object comparison but if still think we should compare JSON, I have now objections. @tmetzke WDYT?

runtime-job-worker/src/test/java/io/camunda/connector/runtime/jobworker/ConnectorJobHandlerTest.java
igpetrov: Fair point.

runtime-job-worker/src/test/java/io/camunda/connector/runtime/jobworker/FeelEngineWrapperTest.java
tmetzke: I see what you mean. The thing is, the expected result ideally shouldn't change when we adjust any code. I would prefer easy-to-read and therefore usually also more maintainable test cases. Potential future sloppy development practices shouldn't be an argument for keeping more complex code that's harder to change, I think. 🙂

igpetrov: I brought the argument regarding developer being lazy because I saw production incidents related to that :). But okay, as I say, I don't have objections in this relation because I in my opinion both make sense to me thus I wil change this.

runtime-job-worker/src/test/java/io/camunda/connector/runtime/jobworker/ConnectorJobHandlerTest.java
```diff
@@ -88,20 +86,128 @@ public void shouldNotSetWithoutResultVariable() throws Exception {
...
+      // given
+      // Response from service -> empty
+      var jobHandler = new ConnectorJobHandler((context) -> new HashMap<>());
+
+      // FEEL expression -> {"processedOutput":response.callStatus}
+      final String resultExpression = "{\"processedOutput\": response.callStatus }";
+
+      // when
+      Assertions.assertThrows(
+          NullPointerException.class,
```
nikku, line 208: Is this the same behavior used in the connector bridge? In the spirit of FEEL (being null-safe, end-to-end) we could revise this behavior at some point.

runtime-job-worker/src/test/java/io/camunda/connector/runtime/jobworker/feel/FeelEngineWrapperTest.java
```diff
@@ -0,0 +1,135 @@
...
+    final String evaluatedResultAsJson =
+        objectUnderTest.evaluateToJson(resultExpression, variables);
+
+    // then
+    assertThat(evaluatedResultAsJson)
+        .isEqualTo("{\"processedOutput\":{\"statusCode\":\"200 OK\"}}");
+  }
+
+  @Test
+  void evaluateFeelExpressionToJson_ShouldSucceed_WhenVariableNotFound() {
```
nikku, line 70: Consider using a nested spec to group `evaluateFeelExpresisonToJSON` tests.

nikku:
Great work. Looks awesome :)

runtime-job-worker/src/main/java/io/camunda/connector/runtime/jobworker/feel/FeelEngineWrapper.java
```diff
@@ -0,0 +1,105 @@
...
+  }
+
+  @SuppressWarnings("unchecked")
+  private static Map<String, Object> ensureVariablesMap(final Object variables) {
+    return (Map<String, Object>)
+        Objects.requireNonNull(variables, ERROR_VARIABLES_MUST_NOT_BE_NULL);
+  }
+
+  public String evaluateToJson(final String expression, final Object variables) {
+    return Optional.of(ensureVariablesMap(variables))
```
tmetzke, line 76: There are two things that strike me here:

1. This line can throw a `ClassCastException` when the input is not a map (but a simple String, for example) or a `NullPointerException` when the input is `null`. Did we test for both scenarios? This should ideally bubble up as a `FeelEngineWrapperException` as it does in the Connector Bridge (see [related test here](https://github.com/camunda/cloud-connector-bridge/blob/7feeefa7ca6845c5decd4a92fcbac9fdfee0173f/src/test/java/io/camunda/connectorbridge/zeebe/FeelEngineWrapperTest.java#L98)). I believe it will currently bubble up as the original exception, which is not that nice and also a bit inconsistent then.
2. The result of `ensureVariablesMap` is never `null`, right? Does it make sense to use `Optional` here? I know it's nice to use the `map` functions and such, but maybe we can even make this code simpler here without all the `Optional` code around. Just a thought.

runtime-job-worker/src/main/java/io/camunda/connector/runtime/jobworker/feel/FeelEngineWrapper.java
```diff
@@ -0,0 +1,105 @@
...
+        .map(
+            context -> {
+              final var result = feelEngine.evalExpression(trimExpression(expression), context);
+              if (result.isRight()) {
+                return resultToJson(result.right().get(), expression, context);
+              } else {
+                throw new FeelEngineWrapperException(
+                    ERROR_EXPRESSION_EVALUATION_FAILED,
+                    expression,
+                    context,
```
tmetzke, line 87: I think we should use the original `variables` as input for the exception. We should state what the original preconditions were for this method call. The `context` is something we already created internally.

runtime-job-worker/src/main/java/io/camunda/connector/runtime/jobworker/feel/FeelEngineWrapper.java
```diff
@@ -0,0 +1,105 @@
...
+
+  private String resultToJson(
+      final Object result,
+      String expression,
+      final scala.collection.immutable.Map<String, Object> context) {
+    try {
+      return objectMapper.writeValueAsString(result);
+    } catch (final JsonProcessingException e) {
+      throw new FeelEngineWrapperException(
+          ERROR_RESULT_CANNOT_BE_PARSED_AS_JSON, expression, context, e);
```
tmetzke, line 102: I think this can simply throw a `RuntimeException`. It should be caught by the caller, mapping this to a `FeelEngineWrapperException` there.

runtime-job-worker/src/main/java/io/camunda/connector/runtime/jobworker/ConnectorJobHandler.java
```diff
@@ -74,10 +82,28 @@ public void handle(final JobClient client, final ActivatedJob job) {
...
+        .ifPresent(map -> outputVariables.putAll(map));
+
+    return outputVariables;
+  }
+
+  private Map<String, Object> parseJsonVarsAsMapOrThrow(final String jsonVars) {
+    try {
+      return OBJECT_MAPPER.readValue(jsonVars, Map.class);
+    } catch (JsonProcessingException e) {
+      throw new RuntimeException(e);
```
tmetzke, line 105: Can we throw a `FeelEngineWrapperException` here for consistency? The API for the `createOutputVariables` should be to either succeed or throw this one exception providing proper context.

runtime-job-worker/src/test/java/io/camunda/connector/runtime/jobworker/ConnectorJobHandlerTest.java
tmetzke: Same general behavior in the bridge, yes. It throws an exception when the map is empty for example.

runtime-job-worker/src/test/java/io/camunda/connector/runtime/jobworker/ConnectorJobHandlerTest.java
```diff
@@ -88,20 +86,128 @@ public void shouldNotSetWithoutResultVariable() throws Exception {
...
+          JobBuilder.create().withResultVariableHeader(resultVariableName).execute(jobHandler);
+
+      // then
+      assertThat(result.getVariables()).containsKey(resultVariableName);
+      assertThat(result.getVariable(resultVariableName)).isNull();
+      assertThat(result.getVariables().size()).isEqualTo(1);
+    }
+
+    @Test
+    public void shouldSetResultVariableNullWhenCallReturnedEmpty() {
```
tmetzke, line 166:
```suggestion
    public void shouldSetResultVariableEmptyWhenCallReturnedEmpty() {
```

runtime-job-worker/src/test/java/io/camunda/connector/runtime/jobworker/ConnectorJobHandlerTest.java
tmetzke: The Connector Bridge throws an exception if the returned connector result value is `null` and an expression is defined on it. When the connector result value is an empty map (as tested here), the Bridge only throws an exception if the result expression references a value that does not exist (this of course also happens when there is data in the variables and I reference something in my expression that is not present). So, the `NullPointerException` here is somewhat inconsistent, I would say.

runtime-job-worker/src/test/java/io/camunda/connector/runtime/jobworker/feel/FeelEngineWrapperTest.java
tmetzke: Since you renamed the evaluate method, we can also rename the test method names here, I believe. Creating a nested spec group might also make sense here, yes. You could also rename the whole test class accordingly as well since we only have expression evaluation tests here. If there are other things in the wrapper to unit test, then different groups might also be nice here in this class.

runtime-job-worker/src/test/java/io/camunda/connector/runtime/jobworker/ConnectorJobHandlerTest.java
igpetrov: The NPE is coming from here: https://github.com/camunda/cloud-connector-bridge/blob/main/src/main/java/io/camunda/connectorbridge/zeebe/FeelEngineWrapper.java#L45

runtime-job-worker/src/test/java/io/camunda/connector/runtime/jobworker/feel/FeelEngineWrapperTest.java
igpetrov: Renamed tests (in the next PR). The testing names are composed from `methodName_ExpectedResult_Conditions`.

runtime-job-worker/src/main/java/io/camunda/connector/runtime/jobworker/ConnectorJobHandler.java
igpetrov: Updated

runtime-job-worker/src/main/java/io/camunda/connector/runtime/jobworker/feel/FeelEngineWrapper.java
igpetrov: Updated

igpetrov:
Updated the "low hanging" comments. Still working on the remaining.

runtime-job-worker/src/main/java/io/camunda/connector/runtime/jobworker/feel/FeelEngineWrapper.java
igpetrov: 1. So the `variables` are coming from top-level response `Object result = call.execute(new JobHandlerContext(job));`. I am not very familiar with FEEL, but I thought if we try to map non-JSON value, it should fail anyway. I like the idea though, to converge to `FeelEngineWrapperException`, maybe with `try-catch` or smth.
2. Correct, I am using it because it's comfortable to use `map-reduce` here. `Optional.of` still throws an `NPE` compared to `Optional.ofNullable`. Let me see if it looks better.

runtime-job-worker/src/test/java/io/camunda/connector/runtime/jobworker/ConnectorJobHandlerTest.java
igpetrov: After refactoring, the FEEL wrapper class only produces FEEL Exception type from public method call.

runtime-job-worker/src/test/java/io/camunda/connector/runtime/jobworker/feel/FeelEngineWrapperTest.java
igpetrov: Renamed test class as well.

runtime-job-worker/src/main/java/io/camunda/connector/runtime/jobworker/feel/FeelEngineWrapper.java
igpetrov: Refactored a bit map + added a test case when `variables` is not a map. Case when `variables` were `null` was present before.

runtime-job-worker/src/main/java/io/camunda/connector/runtime/jobworker/feel/FeelEngineWrapper.java
igpetrov: Done

igpetrov:
Hi @tmetzke @nikku.

Update the PR per request.

nikku:
@tmetzke Please have another look and give this your final approval. 

I'm ready to steal approved and merged code for my inbound prototype.

tmetzke:
On my list for today until 15:30 CEST.

runtime-job-worker/src/main/java/io/camunda/connector/runtime/jobworker/feel/FeelEngineWrapper.java
```diff
@@ -0,0 +1,96 @@
...
+      var variablesAsMapAsScalaMap = toScalaMap(variablesAsMap);
+      var result = feelEngine.evalExpression(trimExpression(expression), variablesAsMapAsScalaMap);
+      if (result.isRight()) {
+        return resultToJson(result.right().get());
+      } else {
+        throw new RuntimeException(result.left().get().message());
+      }
+    } catch (Exception e) {
+      throw new FeelEngineWrapperException(
+          ERROR_EXPRESSION_EVALUATION_FAILED, expression, Objects.toString(variables), e);
```
tmetzke, line 85: Do we need the `Objects#toString` here? We are passing on an object to the exception that will transformed into a String over there with `String.format(Failed to evaluate expression '%s' and context '%s', because %s", expression, context, reason)`, right?
```suggestion
          ERROR_EXPRESSION_EVALUATION_FAILED, expression, variables, e);
```

igpetrov: Good point. I confused this method with more advanced equivalent from other library.

runtime-job-worker/src/test/java/io/camunda/connector/runtime/jobworker/ConnectorJobHandlerTest.java
tmetzke: > After refactoring, the FEEL wrapper class only produces FEEL Exception type from public method call.

But isn't this still throwing an NPE now? Why is that?

tmetzke (commented):
Only one test case (see [open conversation](https://github.com/camunda/connector-sdk/pull/66/#discussion_r955779063)) left to attend. If that one's fixed, I think we're good to go. Well done, @igpetrov! 🎉

runtime-job-worker/src/test/java/io/camunda/connector/runtime/jobworker/ConnectorJobHandlerTest.java
igpetrov: Good point. This is my understanding as well. Looking at it now.

igpetrov: So, after all, it seems it doesn't need to raise an exception, because
```
public void handle(final JobClient client, final ActivatedJob job) {
...
    try {
...
    } catch (Exception error) {
...
      client.newFailCommand(job).retries(0).errorMessage(error.getMessage()).send().join();
    }
  }
```

tmetzke: Thanks for looking into this. I would prefer being able to still assert the exception in the test. Do you think that's possible to achieve here?

tmetzke:
Looks good to me overall, well done 👍 
@igpetrov, there is one last question in the [open conversation](https://github.com/camunda/connector-sdk/pull/66/#discussion_r955779063), maybe you could turn to that to finalize this? If that's not easy to achieve, please create a follow-up and merge this one here. If it's simple enough to do, please attend it with this PR here.

Thank you! 🎉

igpetrov:
> Looks good to me overall, well done 👍 @igpetrov, there is one last question in the [open conversation](https://github.com/camunda/connector-sdk/pull/66/#discussion_r955779063), maybe you could turn to that to finalize this? If that's not easy to achieve, please create a follow-up and merge this one here. If it's simple enough to do, please attend it with this PR here.
> 
> Thank you! 🎉

@tmetzke sorry, I somehow missed that last comment of yours. I'll take a look and see what can be done and let you know later (today?).

tmetzke:
> @tmetzke sorry, I somehow missed that last comment of yours. I'll take a look and see what can be done and let you know later (today?).

Sure, @igpetrov, whenever you find the time 👍

igpetrov:
> > @tmetzke sorry, I somehow missed that last comment of yours. I'll take a look and see what can be done and let you know later (today?).
> 
> Sure, @igpetrov, whenever you find the time 👍

Hi @tmetzke,

After looking into this, the only way to test exception thrown here is via intercepting log messages which doesn't look solid for SDK-tier library and it also requires to include additional libraries.

The other alternative is to decompose existing tests which is preferred from my point of view but also time consuming.

My suggestion here, if this is not critical for now, we merge this code and create a follow up task. WDYT?

runtime-job-worker/src/test/java/io/camunda/connector/runtime/jobworker/ConnectorJobHandlerTest.java
tmetzke: Let's move this into a follow up.

tmetzke:
> My suggestion here, if this is not critical for now, we merge this code and create a follow up task. WDYT?

Yes, let's do that 👍

runtime-job-worker/src/test/java/io/camunda/connector/runtime/jobworker/ConnectorJobHandlerTest.java
igpetrov: Created a ticket https://github.com/camunda/connector-sdk/issues/73

igpetrov:
> > My suggestion here, if this is not critical for now, we merge this code and create a follow up task. WDYT?
> 
> Yes, let's do that 👍

Created a ticket https://github.com/camunda/connector-sdk/issues/73.

Could you please approve this one?

tmetzke (approved):
Thanks, @igpetrov ! 👍

PR #69, opened 2022-08-31
Enable annotation-based validation & secret replacement

### LINKED ISSUE #71: Enable declarative secret replacement configuration for inputs

Configuring secrets replacement in input objects is quite verbose and explicit (implement `ConnectorInput` interface, override `replaceSecrets` and replace secrets in supported attributes explicitly). I have to do this for the whole hierarchy of input objects if there are nested objects.

I would like to annotate the attributes that support secrets.

For example:
```java
record Authentication(String user, @Secret String token) { }
```

Keep using the interface and its method.

__Additional context__

n/a

jonathanlukas:
## Description

The validation was bridged and secret replacement mechanisms were replaced by a declarative way.

The validation can now be bridged to any kind of framework or implementation.

## Related issues

Related to #70
Closes #71


tmetzke:
Hi @jonathanlukas,

thanks for the extensive proposal. Before we dive into discussing concrete code, I'd like to discuss implications and general directions.

Introducing this will have the following implications, to my understanding (please correct me if I'm wrong):

* There are two approaches (combinable) for input validation:
   1. Use validation annotations on parameters (new).
   2. Program validations in the input's `validateWith` method.
* We can have nested input objects (an attribute in an input pointing to another input class with parameters to validate). Do we support this here already with the current setup?
* Strictly speaking, implementing the `ConnectorInput` interface would not be necessary anymore for validation, you can do that with the annotations. However, in order to initiate the validation, you would currently still need to do this as the `ConnectorContext#validate` requires this interface. This might become confusing. Especially if we, for example, move to secret replacement support via annotation as well (which will be a lot easier than validation since it's ONE annotation along the lines of `@Secret` or similar).
* Assuming we remove the current validation mechanism and replace it with annotation-based validation, the only way to write custom validation logic is by providing custom interfaces and `FieldValidationVisitor`s and making them available as SPIs. This might be too big of a hurdle for some people (no evidence, just a gut feeling). Currently, you could simply add the custom validation in the input's `validateWith` method and decide on your own where this validation logic comes from (an external library, a local utility class, local one-time logic in the `validateWith` method, ...).

In general, I like the extensibility of this approach. However, I am not sure yet if we should

1. Support annotation-based AND `validateWith`-based validation together.
1. Only go for annotation-based (and remove `validateWith`-based) validation.
1. Stay with `validateWith`-based validation (and reject annotation-based).

Some factors that I think can influence this:

* If writing custom validation logic would become a bit simpler to achieve with the annotations, I could imagine going for annotation-based validation only.
* Having annotation and `validateWith` side by side (and combinable) might simply be fine (depends on proper documentation, probably). Happy for any additional thoughts here.
* Ideally, allowing secret replacement via annotations should be supported as well if we go for annotations (so you can either skip implementing additional methods in inputs or remove the need for implementing the `ConnectorInput` interface altogether).
* We could keep both validation approaches and remove the extensibility part of the annotations. The default recommendation could then be to 1) use annotations for supported validations (covering >=80% of validation demand) and 2) write custom validation using `validateWith` for more control. We could also keep the extensibility part of the annotations and say 3) provide custom validation annotations. But the story for what to use at which point might become too complex to allow for an easy getting-started approach. Again, just a gut feeling, no evidence here.

Let us know what you think about those points.

jonathanlukas:
> We can have nested input objects (an attribute in an input pointing to another input class with parameters to validate). Do we support this here already with the current setup?

No, this is not yet part of the implementation. I could think of `@NestedInput` as marker.

> Strictly speaking, implementing the ConnectorInput interface would not be necessary anymore for validation, you can do that with the annotations. However, in order to initiate the validation, you would currently still need to do this as the ConnectorContext#validate requires this interface.

True, this is something I was wrapping my head around as well. After all, I came to the conclusion that 2 ways might provide an easy start AND a possible way to scale up complexity. But maybe a input-field wise validation is already an easy start...

> Especially if we, for example, move to secret replacement support via annotation as well (which will be a lot easier than validation since it's ONE annotation along the lines of @Secret or similar).

I have this also already prepared. I wanted to check how open you are for declarative stuff.

> If writing custom validation logic would become a bit simpler to achieve with the annotations, I could imagine going for annotation-based validation only.

Currently, there are 2 things to do:
* create your validation logic - in the easiest case, you would implement the interface straight away. If you want a more generic way, custom annotation creation will be required, meaning this will be 2 steps.
* register is as spi

One way to make it easier would be to create a custom validation annotation which has a class as value which extends an interface that would then be invoked by a FieldValidationVisitor.

> Having annotation and validateWith side by side (and combinable) might simply be fine (depends on proper documentation, probably). Happy for any additional thoughts here.

I am not completely happy with the concept of `validateWith`. Reason is that I prefer separation of logic and data in my code. This concept would mix things up.

> We could keep both validation approaches and remove the extensibility part of the annotations. The default recommendation could then be to 1) use annotations for supported validations (covering >=80% of validation demand) and 2) write custom validation using validateWith for more control. We could also keep the extensibility part of the annotations and say 3) provide custom validation annotations. But the story for what to use at which point might become too complex to allow for an easy getting-started approach. Again, just a gut feeling, no evidence here.

This could work. Anyway, extending it would probably feel like extending some core functionality. So maybe the approach with a custom validation could make things easier.

nikku:
I'm a fan of getting rid of `ConnectorInput` (again). I also like the validation based on annotations: https://camunda.slack.com/archives/C02JLRNQQ05/p1661948829350449?thread_ts=1661935254.284599&cid=C02JLRNQQ05.

Explicitly marking places that may contain secrets could also be a nice benefit. However it does not solve 100% of our issues. I.e. users reported that they'd like to replace secrets in parts of their content i.e. an `Authorization: token {{ secrets.SOME_SECRET }}` header.


The general question is: 

* Shall we go the annotation based approach over the "explicit API" based approach?

nikku:
@tmetzke I think we should settle down on one approach and go that, end-to-end. Or at least not expose some features via our public API that _may_ not be the way to go in the future.

That is just my two cents, of course :).

tmetzke:
@jonathanlukas and I aligned on the following in a call:

* We go for annotation-based validation (and secret replacement, which comes as a follow-up) only.
* We remove `ConnectorInput` and related code.
* We move validation logic into an abstract implementation of `ConnectorContext`, reused by all current implementations (SaaS env, job worker env, test env).
* We cover nested and polymorphic input objects with the annotations
* We provide a validation annotation that takes a class as an argument that allows for custom validation (parameterless constructor only)

@jonathanlukas, please add here if I forgot anything. 🙂

jonathanlukas:
Now I will have to introduce secrets by annotations in this PR as well...

I could not completely remove `ConnectorInput` without introducing a new strategy for handling secret replacement.

Now we have:
* `@NotEmpty`, `@MinLength`, `@MaxLength` and `@Pattern` as default validation annotations (on fields only)
* `@CustomFieldValidation` and `@CustomObjectValidation` as easy way of extending validation
* `@Condition` and `@Id` for conditional fields. They will not be checked if the condition is not met
* `@Secret` as secret replacement marker
* `@NestedProperties` as marker to also validate the nested object
* `AbstractConnectorContext` that required an instance of `ValidationContext`. It has a default implementation to provide a `SecretContext` that can be overridden
* `CustomFieldValidator` and `CustomObjectValidator` interfaces that are referenced from the  `@CustomFieldValidation` and `@CustomObjectValidation` annotations
* `FieldValidationVisitor` interface that is applied on each field of an object to be validated. These visitors are registered in the `AbstractValidationContext`. By extending it, custom `FieldValidationVisitor` instances may be added as well
* `SecretContext` interface that the secret replacement can be delegated to. There is only one implementation that is standalone and is used by default
* `ObjectProviderContext` is a facade for some kind of BeanFactory that returns an object for the given class. The default is to use the default constructor by reflection. For runtime, it is implemented and uses the serviceloader first and reflection as Fallback.

nikku:
As I mentioned to @tmetzke [here](https://github.com/camunda/cloud-connectors/issues/96#issuecomment-1233848462) I'd want us to answer a number of questions before switching our validation approach.

jonathanlukas:
@tmetzke @nikku I implemented what we discussed yesterday. Any thoughts? Please also look [here](https://github.com/camunda/cloud-connectors/issues/96#issuecomment-1235104001)

tmetzke:
I think this goes in the right direction. 👍 

But I would want to simplify this a bit. We need to stay pragmatic about the things that should be extensible/replaceable by custom implementations. Otherwise, we overcomplicate things and make it harder to contribute to the SDK for now, I think. In the beginning, things like secret replacement should be fixed by the SDK by default (you can still override the `replaceSecrets` in a custom context if you need to). Therefore, I would not introduce the following interfaces right now:

* NestedPropertiesDetector
* ReflectionHelper
* SecretContext

I would consolidate the concrete implementations of those interfaces high up in the hierarchy for now (`AbstractConnectorContext` mostly). That way, we can easily reuse this in our environments (test, job worker, cloud). SDK users also get those things out of the box when building their custom runtime environments.

jonathanlukas:
@tmetzke This is also done.

runtime-job-worker/src/main/java/io/camunda/connector/runtime/jobworker/ConnectorJobHandler.java
```diff
@@ -119,5 +112,17 @@ public <T> T getVariablesAsType(Class<T> cls) {
...
+    protected ValidationProvider getValidationProvider() {
+      return ServiceLoader.load(ValidationProvider.class)
+          .findFirst()
+          .orElseThrow(
+              () ->
+                  new IllegalStateException(
+                      "Please bind an implementation to "
+                          + ValidationProvider.class.getName()
+                          + " via SPI"));
+    }
```
tmetzke, line 126: I think we can move this to `AbstractConnectorContext`. That way, the test context can also use it. 

Benefits:

* We can provide a unit test actually that ensures that the SPI lookup works and I can validate my input. It doesn't have to be by annotation. It can be simply a mockup validation provider that does nothing when the validate method is called.
* SDK users can test validation easily as well using the test context class if they have the `connector-validation` artifact in their project. The test context will look up the SPI and use the Java Bean Validation.

WDYT?

core/src/test/java/io/camunda/connector/test/ConnectorFunctionTest.java
```diff
@@ -56,19 +56,6 @@ public void shouldReplaceSecret() throws Exception {
...
-    var fn = new ExampleFunction();
-    var context = ConnectorContextBuilder.create().variables(new ExampleInput()).build();
-
-    // when
-    var exception = catchException(() -> fn.execute(context));
-
-    // then
-    assertThat(exception.getMessage()).contains("Test - foo");
-  }
-
```
tmetzke, line 71: I think we can keep a test case regarding validation that ensures the SPI loading works fine. Plus one that ensures an exception is thrown when there is no SPI.

core/src/main/java/io/camunda/connector/test/ConnectorContextBuilder.java
```diff
@@ -104,10 +105,16 @@ public ConnectorContextBuilder secrets(SecretProvider secretProvider) {
...
+    return new AbstractConnectorContext() {
 
       private SecretStore secretStore;
 
+      @Override
+      protected ValidationProvider getValidationProvider() {
+        // TODO make this more valid maybe
+        return objectToValidate -> {};
+      }
+
```
tmetzke, line 117: We can remove this and move the logic to the `AbstractConnectorContext`. See related comment in `ConnectorJobHandler`.

core/src/main/java/io/camunda/connector/impl/AbstractConnectorContext.java
```diff
@@ -0,0 +1,127 @@
...
+              .getDeclaredMethod(
+                  "set"
+                      + field.getName().substring(0, 1).toUpperCase()
+                      + field.getName().substring(1),
+                  field.getType());
+      setter.invoke(containingObject, property);
+    } catch (Exception e) {
+      throw new RuntimeException(e);
+    }
+  }
```
tmetzke, line 111: Maybe we can have a look at how this is done in a library like https://github.com/hibernate/hibernate-validator. Just for reference. If you already did this, that's fine. But I think the Java bean standard defines that for booleans the getter is named `isXYZ()`, for example.

core/src/main/java/io/camunda/connector/impl/AbstractConnectorContext.java
```diff
@@ -0,0 +1,127 @@
...
+          .peek(
+              field -> {
+                Object property = getProperty(input, field);
+                if (hasNestedProperties(field, property)) {
+                  replaceSecrets(property);
+                }
+              })
+          .filter(field -> field.isAnnotationPresent(Secret.class))
+          .forEach(field -> handleSecret(input, field));
+    }
```
tmetzke, line 53: This looks generally fine to me. Let's add some test cases that ensure we covered the cases we want to support.

tmetzke (commented):
Looks really good already, regarding the SDK Core. Conceptually, I think we're there 👍 
I added some suggestions regarding context and tests.

runtime-job-worker/src/main/java/io/camunda/connector/runtime/jobworker/ConnectorJobHandler.java
jonathanlukas: I was unsure as in no other place (from what I could see), the core component tried to load something from the serviceloader.

tmetzke: OK, I see. Let's see if we need to move this there:

* The Connector runtime environment can decide how to find a validation provider. 
* Connector function developers can include the provided `connector-validation` artifact, not knowing where the function will run since it is environment-agnostic. 
* A Connector runtime environment needs to support the loading mechanism provided by `connector-validation` out of the box, otherwise, it can not use all available connector functions out there.
* If we don't integrate this loading mechanism by default (or make it easy to integrate without writing it yourself), environment developers will potentially struggle.

Based on this, I would argue to move this to the core. People can still override it for their environment's context if they don't need this (because they don't need validation, bring along their own in their environment, or whatever other reason).

jonathanlukas: I see. As of now, we did not have to integrate "horizontally", this was not required in core. Now, this is the default integration mechanism for horizontal module interaction.

tmetzke: Open for any other ideas that would make sense here 🙂

jonathanlukas: Nah it is fine. I will move this upwards for seemless integration.

tmetzke: I had another quick look as well, Service Loader is by far the most common approach if you don't use Spring or something similar directly.

core/src/main/java/io/camunda/connector/impl/AbstractConnectorContext.java
jonathanlukas: I looked into hibernates' Reflection Helper which is much more basic as our use case it quite specific. Anyway, these methods could be overridden to bridge to another reflection framework.

I also added some tests to assert behaviour for each field access modifier.

core/src/main/java/io/camunda/connector/impl/AbstractConnectorContext.java
jonathanlukas: Done. Now, there is a test that is replacing secrets inside arrays and stuff.

core/src/test/java/io/camunda/connector/test/ConnectorFunctionTest.java
jonathanlukas: There is a "shouldFail" test function that asserts the attempted usage of the ValidationProvider

core/src/main/java/io/camunda/connector/impl/AbstractConnectorContext.java
tmetzke: I am not all that certain if we should by default scan ALL fields of an object and look for nested `@Secret` annotations automatically. That might become a performance issue. I think we should only consider those container fields (arrays, iterables, ...) that are annotated with `@Secret` as well (or another annotation if reusing that annotation doesn't work). It's similar to how `@Valid` works than, for example.

core/src/test/java/io/camunda/connector/impl/ComplexTestInput.java
```diff
@@ -0,0 +1,24 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.camunda.connector.impl;
+
+import java.util.Set;
+
+public class ComplexTestInput {
+  public TestInput[] inputArray = new TestInput[] {new TestInput(), new TestInput()};
+  public Set<TestInput> testInputs = Set.of(new TestInput(), new TestInput());
```
tmetzke, line 23: I think it should be necessary to annotate those container fields that should be checked for secrets. Let's not do this automatically for ALL fields in an object.

core/src/test/java/io/camunda/connector/impl/ConnectorContextTest.java
```diff
@@ -0,0 +1,118 @@
...
+import org.junit.jupiter.api.Test;
+
+public class ConnectorContextTest {
+  private static Field fieldForName(Class<?> clazz, String fieldName) throws NoSuchFieldException {
+    return clazz.getDeclaredField(fieldName);
+  }
+
+  @Test
+  public void shouldGetPropertyFromDeclaredFields() throws NoSuchFieldException {
+    TestConnectorContext connectorContext = new TestConnectorContext();
```
tmetzke, line 33: Is there a specific reason NOT to use `ConnectorContextBuilder` here? If so, we should adjust this helper class. If it's not useable enough for us, it won't be for others as well.

core/src/test/java/io/camunda/connector/test/ConnectorFunctionTest.java
```diff
@@ -56,29 +55,19 @@ public void shouldReplaceSecret() throws Exception {
...
   @Test
   public void shouldFail() {
```
tmetzke, line 59: This test case ensures we can fail the logic of a Connector in an expected way. I would prefer to have a separate failing test case for a missing validation provider. Plus, we should also add a test case for a provided validation provider (doesn't have to be a complex provider).

core/src/test/java/io/camunda/connector/impl/ConnectorContextTest.java
```diff
@@ -0,0 +1,118 @@
...
+    assertEquals(modified, testInput.getPrivateField());
+    assertThrows(
+        IllegalStateException.class,
+        () ->
+            connectorContext.setProperty(
+                testInput, fieldForName(TestInput.class, "finalField"), modified));
+    connectorContext.setProperty(
+        testInput, fieldForName(TestInput.class, "packageField"), modified);
+    assertEquals(modified, testInput.packageField);
+  }
```
tmetzke, line 73: Let's try to write atomic test cases. One test case should test one method call in a `// when` block. Please also try to use `given - when - then` blocks in the test cases. 😃 See also [good unit tests guide](https://gist.github.com/nikku/1dc2edb553565238c7bf).

core/src/test/java/io/camunda/connector/impl/ComplexTestInput.java
jonathanlukas: I think this has a dependency towards the validation framework used as I do not want to add 2 annotations to check for nested secrets and nested validations seperately.

core/src/test/java/io/camunda/connector/impl/ConnectorContextTest.java
jonathanlukas: I will extend `ConnectorContextBuilder` instead. I have to override the `getProperty` and `setProperty` privacy.

core/src/test/java/io/camunda/connector/test/ConnectorFunctionTest.java
jonathanlukas: I added it and enhanced the `ConnectorContextBuilder` to build it with a `ValidationProvider` on the fly.

core/src/test/java/io/camunda/connector/impl/ConnectorContextTest.java
jonathanlukas: Gotcha. Missed out on that basic....

core/src/main/java/io/camunda/connector/impl/AbstractConnectorContext.java
tmetzke: That is actually what I would not do by default. I think the array and Iterable parts are fine. For the `else` branch, I'd probably only check those fields for nested properties that have the `@Secret` annotation (if we use that for marking nested secret containers as well).

core/src/test/java/io/camunda/connector/impl/ComplexTestInput.java
tmetzke: I think the validation framework handles nested properties itself via `@Valid`. I believe we can restrict looking for nested secret properties by using `@Secret` on those object properties. No need to have a dependency on the validation framework.

core/src/main/java/io/camunda/connector/impl/AbstractConnectorContext.java
tmetzke: Example:

```java
public class MyConnectorInput {

  @Secret
  private String url;

  @Secret
  private Authentication authentication;

  private ComplexNestedObject otherProperty;
  ...
}

public class Authentication {

  @Secret
  private String token;

  private String type;
}
```

I would not want `otherProperty` and `type` to be examined in the secret replacement

tmetzke:
@jonathanlukas, can we for now only make additional commits instead of rebasing? This will make reviewing a lot easier as I can track incremental changes faster. We can squash at one point when we're settled on all issues. Is that fine with you?

core/src/test/java/io/camunda/connector/impl/ComplexTestInput.java
tmetzke:
```suggestion
  @Secret
  public TestInput[] inputArray = new TestInput[] {new TestInput(), new TestInput()};

  @Secret
  public Set<TestInput> testInputs = Set.of(new TestInput(), new TestInput());
```

tmetzke: Plus, we should make this more complex to cover more use cases, I think. We should involve other objects that can have two simple and two nested properties again, one of each type annotated with `@Secret`. Just to check that our graph resolution behaves as expected.

core/src/test/java/io/camunda/connector/impl/ConnectorContextTest.java
```diff
@@ -0,0 +1,127 @@
...
+            new GetPropertyTestInput("finalField", testInput.finalField),
+            new GetPropertyTestInput("packageField", testInput.packageField));
+    ThrowingConsumer<GetPropertyTestInput> executor =
+        in -> {
+          String content =
+              connectorContext.getProperty(testInput, fieldForName(TestInput.class, in.fieldName));
+          assertEquals(in.expectedContent, content);
+        };
+    return DynamicTest.stream(input, i -> i.fieldName, executor);
+  }
```
tmetzke, line 58:
```suggestion
  public Stream<DynamicTest> shouldGetPropertyFromField() {
    // given
    TestConnectorContext connectorContext = ConnectorContextBuilder.create().build();
    TestInput testInput = new TestInput();
    Stream<GetPropertyTestInput> input =
        Stream.of(
            new GetPropertyTestInput("publicField", testInput.publicField),
            new GetPropertyTestInput("protectedField", testInput.protectedField),
            new GetPropertyTestInput("privateField", testInput.getPrivateField()),
            new GetPropertyTestInput("finalField", testInput.finalField),
            new GetPropertyTestInput("packageField", testInput.packageField));
    ThrowingConsumer<GetPropertyTestInput> executor =
        in -> {
          // when
          String content =
              connectorContext.getProperty(testInput, fieldForName(TestInput.class, in.fieldName));
          // then
          assertEquals(in.expectedContent, content);
        };
    return DynamicTest.stream(input, i -> i.fieldName, executor);
  }
```

core/src/test/java/io/camunda/connector/impl/ConnectorContextTest.java
```diff
@@ -0,0 +1,127 @@
...
+            new SetPropertyTestInput("privateField", testInput::getPrivateField),
+            new SetPropertyTestInput("packageField", testInput::getPackageField));
+    ThrowingConsumer<SetPropertyTestInput> executor =
+        in -> {
+          connectorContext.setProperty(
+              testInput, fieldForName(TestInput.class, in.fieldName), modified);
+          assertEquals(modified, in.getter.get());
+        };
+    return DynamicTest.stream(input, i -> i.fieldName, executor);
+  }
```
tmetzke, line 78:
```suggestion
  @TestFactory
  public Stream<DynamicTest> shouldSetPropertyToField() {
    // given
    TestConnectorContext connectorContext = ConnectorContextBuilder.create().build();
    TestInput testInput = new TestInput();
    String modified = "modified";
    Stream<SetPropertyTestInput> input =
        Stream.of(
            new SetPropertyTestInput("publicField", testInput::getPublicField),
            new SetPropertyTestInput("protectedField", testInput::getProtectedField),
            new SetPropertyTestInput("privateField", testInput::getPrivateField),
            new SetPropertyTestInput("packageField", testInput::getPackageField));
    ThrowingConsumer<SetPropertyTestInput> executor =
        in -> {
          // when
          connectorContext.setProperty(
              testInput, fieldForName(TestInput.class, in.fieldName), modified);
          // then
          assertEquals(modified, in.getter.get());
        };
    return DynamicTest.stream(input, i -> i.fieldName, executor);
  }
```

core/src/test/java/io/camunda/connector/impl/ConnectorContextTest.java
```diff
@@ -0,0 +1,127 @@
...
+    IllegalStateException exception =
+        assertThrows(
+            IllegalStateException.class,
+            () ->
+                connectorContext.setProperty(
+                    testInput, fieldForName(TestInput.class, "finalField"), modified));
+    assertEquals(
+        "Cannot invoke set or setter on final field 'finalField' of type class io.camunda.connector.impl.TestInput",
+        exception.getMessage());
+  }
```
tmetzke, line 94:
```suggestion
  @Test
  public void shouldThrowWhenSettingOnFinalField() {
    // given
    TestConnectorContext connectorContext = ConnectorContextBuilder.create().build();
    TestInput testInput = new TestInput();
    String modified = "modified";
    IllegalStateException exception =
        assertThrows(
            IllegalStateException.class,
            () ->
                // when
                connectorContext.setProperty(
                    testInput, fieldForName(TestInput.class, "finalField"), modified));
    // then
    assertEquals(
        "Cannot invoke set or setter on final field 'finalField' of type class io.camunda.connector.impl.TestInput",
        exception.getMessage());
  }
```

core/src/test/java/io/camunda/connector/test/ConnectorFunctionTest.java
```diff
@@ -31,7 +30,11 @@ public void shouldExecuteConnector() throws Exception {
     var fn = new ExampleFunction();
 
     // when
-    var context = ConnectorContextBuilder.create().variables(new ExampleInput("FOO")).build();
+    var context =
+        ConnectorContextBuilder.create()
+            .variables(new ExampleInput("FOO"))
+            .validation(input -> {})
```
tmetzke, line 36: It might be easier to adjust the `ExampleFunction` to not call validate by default. We can remove this modification here then. By default, the Connector function don't call validate anymore. Only if they use the (to be built) validation module.

core/src/test/java/io/camunda/connector/test/ConnectorFunctionTest.java
```diff
@@ -48,6 +51,7 @@ public void shouldReplaceSecret() throws Exception {
         ConnectorContextBuilder.create()
             .variables(new ExampleInput("secrets.FOO"))
             .secret("FOO", "SECRET_FOO")
+            .validation(input -> {})
```
tmetzke, line 54: It might be easier to adjust the `ExampleFunction` to not call validate by default. We can remove this modification here then. By default, the Connector function don't call validate anymore. Only if they use the (to be built) validation module.

core/src/test/java/io/camunda/connector/test/ConnectorFunctionTest.java
```diff
@@ -70,15 +81,17 @@ public void shouldValidateInput() {
   }
 
   @Test
-  public void shouldFail() {
+  public void shouldFailOnMissingValidationProvider() {
```
tmetzke, line 84: It might be easier to adjust the `ExampleFunction` to not call validate by default. We can remove this modification here then and leave this test as is. By default, the Connector function don't call validate anymore. Only if they use the (to be built) validation module. We can add another test that uses a connector function that calls validate but doesn't provide a validation provider.

tmetzke (commented):
I like the new context builder and the new tests. 

I think there is still some room for improvement regarding the secret replacement, more precisely the graph resolution it uses. Let's also cover this with a bit more testing to make sure we have good documentation and an understanding of how it behaves in detail.

I think we're good then regarding the SDK side of things, well done, @jonathanlukas 👍 

One question regarding the validation module: Would you like to build this as well? Would like any further input/pointers for that?

jonathanlukas:
> @jonathanlukas, can we for now only make additional commits instead of rebasing? This will make reviewing a lot easier as I can track incremental changes faster. We can squash at one point when we're settled on all issues. Is that fine with you?

It totally is - I can remember working on something else with Nico where he told me that they will only merge PRs containing only 1 commit. I will continue with seperate commits...

core/src/main/java/io/camunda/connector/impl/AbstractConnectorContext.java
jonathanlukas: Got it. So `@Secret` marks the path AND the secret itself. I will change this and create more tests on it.

core/src/test/java/io/camunda/connector/impl/ComplexTestInput.java
jonathanlukas: I have added deep nested secrets (3 layers, can be extended). Only fields marked as `@Secret` are queried for nested properties now.

core/src/test/java/io/camunda/connector/test/ConnectorFunctionTest.java
jonathanlukas: I had this implemented before. I can restore it.

jonathanlukas:
I have worked on the other things as well now. If you like, I can continue with implementing a validation component. Anyway, let's talk before and note down the results here.

core/src/test/java/io/camunda/connector/example/ExampleFunction.java
```diff
@@ -25,13 +25,14 @@ public class ExampleFunction implements ConnectorFunction {
...
 
     var input = context.getVariablesAsType(ExampleInput.class);
-
-    context.validate(input);
+    if ("foo".equals(input.getFoo())) {
+      context.validate(input);
+    }
     context.replaceSecrets(input);
 
     var foo = input.getFoo();
```
tmetzke, line 33: Very minor detail, but can we move this `var foo = input.getFoo();` up a bit so we can reuse the variable? You did not create this, but since we're on it now, we might as well fix this.

core/src/test/java/io/camunda/connector/impl/ConnectorContextTest.java
```diff
@@ -0,0 +1,147 @@
...
+
+  @Test
+  public void shouldReplaceSecretsAlongAnnotations() {
+    // given
+    TestConnectorContext connectorContext =
+        ConnectorContextBuilder.create().secret("geheim", "geheim").build();
```
tmetzke, line 109: Again, a minor thing, but can we avoid the German word here? 😄 
I also kind of like it, but I think as an international SDK we should rather keep this international maybe 🤷

tmetzke (commented):
This looks really good, I like this 👍 
Again, very well done, very constructive discussions 🥇 
I noted some very minor details for the last finishing touch.

tmetzke:
> It totally is - I can remember working on something else with Nico where he told me that they will only merge PRs containing only 1 commit. I will continue with seperate commits...

I think there are two different scenarios:

1. You create a new PR and want the first review. At this state, there should only be one commit (ideally but can also be multiple ones if they are semantically necessary, like a refactoring commit and a new feature commit).
2. You received your first feedback. I personally prefer incremental commits after that until all feedback is done. At that point, you can squash locally and force push with lease or ask the code owner to squash-merge. I think that is both fair.

Always happy for new ideas on how to do this best.

tmetzke:
> @tmetzke , I have worked on the other things as well now. If you like, I can continue with implementing a validation component. Anyway, let's talk before and note down the results here.

Let's discuss the validation module over at #70?

core/src/test/java/io/camunda/connector/example/ExampleFunction.java
jonathanlukas: I removed the var foo completely as the String is no reference but the secret replacement requires the replaced secret as return value.

tmetzke: Good catch, makes sense 👍

tmetzke (approved):
Really nice, thanks @jonathanlukas 👍
This is an awesome contribution 🥇 
Looking forward to the `connector-validation` module 🎉

igpetrov:
Hi @jonathanlukas @tmetzke,

While investigating a failing test at https://github.com/camunda/connector-slack/pull/90, I found a potential bug and would like to validate with both of you.

Given [code snippet in the Slack connector](https://github.com/camunda/connector-slack/blob/main/src/main/java/io/camunda/connector/slack/ChatPostMessageData.java#L71-L76), it generates anonymous lambda function under the hood. See the last method on the screenshot:

Screenshot 2022-09-23 at 19 33 58

This leads to an exception `java.lang.RuntimeException: java.lang.IllegalArgumentException: non-null object for private static boolean io.camunda.connector.slack.ChatPostMessageData.lambda$getUserId$0(java.lang.String,com.slack.api.model.User)`:

<details><summary>Click to expand</summary>

```
java.lang.RuntimeException: java.lang.IllegalArgumentException: non-null object for private static boolean io.camunda.connector.slack.ChatPostMessageData.lambda$getUserId$0(java.lang.String,com.slack.api.model.User)

	at io.camunda.connector.impl.outbound.AbstractOutboundConnectorContext.setProperty(AbstractOutboundConnectorContext.java:234)
	at io.camunda.connector.impl.outbound.AbstractOutboundConnectorContext.lambda$handleSecretsField$4(AbstractOutboundConnectorContext.java:115)
	at io.camunda.connector.impl.outbound.AbstractOutboundConnectorContext.lambda$handleSecretsElement$6(AbstractOutboundConnectorContext.java:127)
	at java.base/java.util.Optional.ifPresent(Optional.java:178)
	at io.camunda.connector.impl.outbound.AbstractOutboundConnectorContext.handleSecretsElement(AbstractOutboundConnectorContext.java:121)
	at io.camunda.connector.impl.outbound.AbstractOutboundConnectorContext.lambda$handleSecretsField$5(AbstractOutboundConnectorContext.java:107)
	at java.base/java.util.stream.ForEachOps$ForEachOp$OfRef.accept(ForEachOps.java:183)
	at java.base/java.util.stream.ReferencePipeline$2$1.accept(ReferencePipeline.java:179)
	at java.base/java.util.Spliterators$ArraySpliterator.forEachRemaining(Spliterators.java:992)
	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:509)
	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:499)
	at java.base/java.util.stream.ForEachOps$ForEachOp.evaluateSequential(ForEachOps.java:150)
	at java.base/java.util.stream.ForEachOps$ForEachOp$OfRef.evaluateSequential(ForEachOps.java:173)
	at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
	at java.base/java.util.stream.ReferencePipeline.forEach(ReferencePipeline.java:596)
	at io.camunda.connector.impl.outbound.AbstractOutboundConnectorContext.handleSecretsField(AbstractOutboundConnectorContext.java:105)
	at io.camunda.connector.impl.outbound.AbstractOutboundConnectorContext.replaceSecrets(AbstractOutboundConnectorContext.java:58)
	at io.camunda.connector.impl.outbound.AbstractOutboundConnectorContext.lambda$handleSecretsElement$6(AbstractOutboundConnectorContext.java:124)
	at java.base/java.util.Optional.ifPresent(Optional.java:178)
	at io.camunda.connector.impl.outbound.AbstractOutboundConnectorContext.handleSecretsElement(AbstractOutboundConnectorContext.java:121)
	at io.camunda.connector.impl.outbound.AbstractOutboundConnectorContext.lambda$handleSecretsField$5(AbstractOutboundConnectorContext.java:107)
	at java.base/java.util.stream.ForEachOps$ForEachOp$OfRef.accept(ForEachOps.java:183)
	at java.base/java.util.stream.ReferencePipeline$2$1.accept(ReferencePipeline.java:179)
	at java.base/java.util.Spliterators$ArraySpliterator.forEachRemaining(Spliterators.java:992)
	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:509)
	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:499)
	at java.base/java.util.stream.ForEachOps$ForEachOp.evaluateSequential(ForEachOps.java:150)
	at java.base/java.util.stream.ForEachOps$ForEachOp$OfRef.evaluateSequential(ForEachOps.java:173)
	at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
	at java.base/java.util.stream.ReferencePipeline.forEach(ReferencePipeline.java:596)
	at io.camunda.connector.impl.outbound.AbstractOutboundConnectorContext.handleSecretsField(AbstractOutboundConnectorContext.java:105)
	at io.camunda.connector.impl.outbound.AbstractOutboundConnectorContext.replaceSecrets(AbstractOutboundConnectorContext.java:58)
	at io.camunda.connector.slack.SlackConnectorRequestTest.replaceSecrets_shouldReplaceSecrets(SlackConnectorRequestTest.java:77)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:568)
	at org.junit.platform.commons.util.ReflectionUtils.invokeMethod(ReflectionUtils.java:727)
	at org.junit.jupiter.engine.execution.MethodInvocation.proceed(MethodInvocation.java:60)
	at org.junit.jupiter.engine.execution.InvocationInterceptorChain$ValidatingInvocation.proceed(InvocationInterceptorChain.java:131)
	at org.junit.jupiter.engine.extension.TimeoutExtension.intercept(TimeoutExtension.java:156)
	at org.junit.jupiter.engine.extension.TimeoutExtension.interceptTestableMethod(TimeoutExtension.java:147)
	at org.junit.jupiter.engine.extension.TimeoutExtension.interceptTestMethod(TimeoutExtension.java:86)
	at org.junit.jupiter.engine.execution.InterceptingExecutableInvoker$ReflectiveInterceptorCall.lambda$ofVoidMethod$0(InterceptingExecutableInvoker.java:103)
	at org.junit.jupiter.engine.execution.InterceptingExecutableInvoker.lambda$invoke$0(InterceptingExecutableInvoker.java:93)
	at org.junit.jupiter.engine.execution.InvocationInterceptorChain$InterceptedInvocation.proceed(InvocationInterceptorChain.java:106)
	at org.junit.jupiter.engine.execution.InvocationInterceptorChain.proceed(InvocationInterceptorChain.java:64)
	at org.junit.jupiter.engine.execution.InvocationInterceptorChain.chainAndInvoke(InvocationInterceptorChain.java:45)
	at org.junit.jupiter.engine.execution.InvocationInterceptorChain.invoke(InvocationInterceptorChain.java:37)
	at org.junit.jupiter.engine.execution.InterceptingExecutableInvoker.invoke(InterceptingExecutableInvoker.java:92)
	at org.junit.jupiter.engine.execution.InterceptingExecutableInvoker.invoke(InterceptingExecutableInvoker.java:86)
	at org.junit.jupiter.engine.descriptor.TestMethodTestDescriptor.lambda$invokeTestMethod$7(TestMethodTestDescriptor.java:217)
	at org.junit.platform.engine.support.hierarchical.ThrowableCollector.execute(ThrowableCollector.java:73)
	at org.junit.jupiter.engine.descriptor.TestMethodTestDescriptor.invokeTestMethod(TestMethodTestDescriptor.java:213)
	at org.junit.jupiter.engine.descriptor.TestMethodTestDescriptor.execute(TestMethodTestDescriptor.java:138)
	at org.junit.jupiter.engine.descriptor.TestMethodTestDescriptor.execute(TestMethodTestDescriptor.java:68)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.lambda$executeRecursively$6(NodeTestTask.java:151)
	at org.junit.platform.engine.support.hierarchical.ThrowableCollector.execute(ThrowableCollector.java:73)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.lambda$executeRecursively$8(NodeTestTask.java:141)
	at org.junit.platform.engine.support.hierarchical.Node.around(Node.java:137)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.lambda$executeRecursively$9(NodeTestTask.java:139)
	at org.junit.platform.engine.support.hierarchical.ThrowableCollector.execute(ThrowableCollector.java:73)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.executeRecursively(NodeTestTask.java:138)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.execute(NodeTestTask.java:95)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1511)
	at org.junit.platform.engine.support.hierarchical.SameThreadHierarchicalTestExecutorService.invokeAll(SameThreadHierarchicalTestExecutorService.java:41)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.lambda$executeRecursively$6(NodeTestTask.java:155)
	at org.junit.platform.engine.support.hierarchical.ThrowableCollector.execute(ThrowableCollector.java:73)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.lambda$executeRecursively$8(NodeTestTask.java:141)
	at org.junit.platform.engine.support.hierarchical.Node.around(Node.java:137)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.lambda$executeRecursively$9(NodeTestTask.java:139)
	at org.junit.platform.engine.support.hierarchical.ThrowableCollector.execute(ThrowableCollector.java:73)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.executeRecursively(NodeTestTask.java:138)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.execute(NodeTestTask.java:95)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1511)
	at org.junit.platform.engine.support.hierarchical.SameThreadHierarchicalTestExecutorService.invokeAll(SameThreadHierarchicalTestExecutorService.java:41)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.lambda$executeRecursively$6(NodeTestTask.java:155)
	at org.junit.platform.engine.support.hierarchical.ThrowableCollector.execute(ThrowableCollector.java:73)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.lambda$executeRecursively$8(NodeTestTask.java:141)
	at org.junit.platform.engine.support.hierarchical.Node.around(Node.java:137)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.lambda$executeRecursively$9(NodeTestTask.java:139)
	at org.junit.platform.engine.support.hierarchical.ThrowableCollector.execute(ThrowableCollector.java:73)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.executeRecursively(NodeTestTask.java:138)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.execute(NodeTestTask.java:95)
	at org.junit.platform.engine.support.hierarchical.SameThreadHierarchicalTestExecutorService.submit(SameThreadHierarchicalTestExecutorService.java:35)
	at org.junit.platform.engine.support.hierarchical.HierarchicalTestExecutor.execute(HierarchicalTestExecutor.java:57)
	at org.junit.platform.engine.support.hierarchical.HierarchicalTestEngine.execute(HierarchicalTestEngine.java:54)
	at org.junit.platform.launcher.core.EngineExecutionOrchestrator.execute(EngineExecutionOrchestrator.java:147)
	at org.junit.platform.launcher.core.EngineExecutionOrchestrator.execute(EngineExecutionOrchestrator.java:127)
	at org.junit.platform.launcher.core.EngineExecutionOrchestrator.execute(EngineExecutionOrchestrator.java:90)
	at org.junit.platform.launcher.core.EngineExecutionOrchestrator.lambda$execute$0(EngineExecutionOrchestrator.java:55)
	at org.junit.platform.launcher.core.EngineExecutionOrchestrator.withInterceptedStreams(EngineExecutionOrchestrator.java:102)
	at org.junit.platform.launcher.core.EngineExecutionOrchestrator.execute(EngineExecutionOrchestrator.java:54)
	at org.junit.platform.launcher.core.DefaultLauncher.execute(DefaultLauncher.java:114)
	at org.junit.platform.launcher.core.DefaultLauncher.execute(DefaultLauncher.java:86)
	at org.junit.platform.launcher.core.DefaultLauncherSession$DelegatingLauncher.execute(DefaultLauncherSession.java:86)
	at org.junit.platform.launcher.core.SessionPerRequestLauncher.execute(SessionPerRequestLauncher.java:53)
	at com.intellij.junit5.JUnit5IdeaTestRunner.startRunnerWithArgs(JUnit5IdeaTestRunner.java:57)
	at com.intellij.rt.junit.IdeaTestRunner$Repeater$1.execute(IdeaTestRunner.java:38)
	at com.intellij.rt.execution.junit.TestsRepeater.repeat(TestsRepeater.java:11)
	at com.intellij.rt.junit.IdeaTestRunner$Repeater.startRunnerWithArgs(IdeaTestRunner.java:35)
	at com.intellij.rt.junit.JUnitStarter.prepareStreamsAndStart(JUnitStarter.java:235)
	at com.intellij.rt.junit.JUnitStarter.main(JUnitStarter.java:54)
Caused by: java.lang.IllegalArgumentException: non-null object for private static boolean io.camunda.connector.slack.ChatPostMessageData.lambda$getUserId$0(java.lang.String,com.slack.api.model.User)
	at java.base/java.lang.reflect.AccessibleObject.canAccess(AccessibleObject.java:454)
	at io.camunda.connector.impl.outbound.AbstractOutboundConnectorContext.lambda$setProperty$13(AbstractOutboundConnectorContext.java:211)
	at java.base/java.util.stream.ReferencePipeline$2$1.accept(ReferencePipeline.java:178)
	at java.base/java.util.Spliterators$ArraySpliterator.tryAdvance(Spliterators.java:1002)
	at java.base/java.util.stream.ReferencePipeline.forEachWithCancel(ReferencePipeline.java:129)
	at java.base/java.util.stream.AbstractPipeline.copyIntoWithCancel(AbstractPipeline.java:527)
	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:513)
	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:499)
	at java.base/java.util.stream.FindOps$FindOp.evaluateSequential(FindOps.java:150)
	at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
	at java.base/java.util.stream.ReferencePipeline.findFirst(ReferencePipeline.java:647)
	at io.camunda.connector.impl.outbound.AbstractOutboundConnectorContext.setProperty(AbstractOutboundConnectorContext.java:224)
	... 102 more
```

</details>

I think, this is a bug. While we can rewrite [the original code snippet](https://github.com/camunda/connector-slack/blob/main/src/main/java/io/camunda/connector/slack/ChatPostMessageData.java#L71-L76) not to use anonymous lambdas, it doesn't solve the issue when other customers may want to use streams or other functional utilities.

Another option is to introduce additional filter statement `.filter(method -> !Modifier.isStatic(method.getModifiers()))` [here](https://github.com/camunda/connector-sdk/blob/main/core/src/main/java/io/camunda/connector/impl/outbound/AbstractOutboundConnectorContext.java#L208) but I am not sure whether we would face with any side-effects.

WDYT?

cc: @Oleksiivanov

PR #74, opened 2022-09-06
feat: support throwing BpmnErrors from Connectors

### LINKED ISSUE #68: Support throwing BPMN Errors from Connectors

In BPMN model, If a task throw an exception, we catch this exception based on error code using error boundary event or an error event sub-process. I was using this feature to start alternate flow based on these error codes.
In Camunda Connector SDK, I have found only one exception class _**ConnectorInputException(Throwable exception)**_.
But I'm unable to use error boundary event to catch this exception.  As  it requires error code.
image

While I was using **Zeebe** to develop Camunda Connectors, I have used  _**ZeebeBpmnError(String errorCode, String errorMessage)**_ to throw exception from Worker class which gets propagated to the BPMN. I would catch this in BPMN and do some conditional operation.
I think we can have a class which can takes error Code as an argument either with or without - errorMessage/ Throwable.

As suggested by @tmetzke https://github.com/camunda/connector-sdk/issues/68#issuecomment-1231191700
### **Please consider these three new capabilities:**
- [x] Support providing an errorCode in an exception in a connector. (#74)
- [x] Evaluate a custom task header errorExpression in the connector runtime environments that invoke the connector functions 
and throw a BPMN Error for the first condition that matches. (#74)
- [ ] #199

As I'm using Connector SDK to develop the connectors, I tried ConnectorInputException class but I'm not able to use it for my purpose of catching error in BPMN based on error code.

__Additional context__

I tried catching error throw from ConnectorInputException in BPMN but it's not working.
In this case, Worker thrown an Invalid username or password exception ( errorCode 123B ).
image
image
image

tmetzke:
## Description

Allows defining an `errorExpression` that configures under which conditions to throw a BpmnError from a Connector.

## Related issues

closes #68


runtime-job-worker/src/main/java/io/camunda/connector/runtime/jobworker/feel/FeelConnectorFunctionProvider.java
```diff
@@ -0,0 +1,63 @@
...
+      Map.of("bpmnError", BPMN_ERROR_FUNCTION);
+
+  @Override
+  public Optional<JavaFunction> resolveFunction(String functionName) {
+    return Optional.ofNullable(functions.get(functionName));
+  }
+
+  @Override
+  public Collection<String> getFunctionNames() {
+    return null;
```
tmetzke, line 53:
```suggestion
    return List.of("bpmnError");
```

runtime-util/src/main/java/io/camunda/connector/runtime/util/feel/FeelConnectorFunctionProvider.java
```diff
@@ -0,0 +1,66 @@
...
+public class FeelConnectorFunctionProvider extends JavaFunctionProvider {
+
+  private static final List<String> ARGUMENTS_BPMN_ERROR = List.of("code", "message");
+  private static final JavaFunction BPMN_ERROR_FUNCTION =
+      new JavaFunction(
+          ARGUMENTS_BPMN_ERROR,
+          args ->
+              new ValContext(
+                  new Context.StaticContext(
+                      new Map.Map2<>("code", toString(args, 0), "message", toString(args, 1)),
```
menski, line 40:
```suggestion
                      new Map.Map2<>(ARGUMENTS_BPMN_ERROR.get(0), toString(args, 0), ARGUMENTS_BPMN_ERROR.get(1), toString(args, 1)),
```

runtime-util/src/main/java/io/camunda/connector/runtime/util/feel/FeelConnectorFunctionProvider.java
```diff
@@ -0,0 +1,66 @@
...
+                      Map$.MODULE$.empty())));
+
+  private static final java.util.Map<String, JavaFunction> functions =
+      java.util.Map.of("bpmnError", BPMN_ERROR_FUNCTION);
```
menski, line 44: `bpmnError` could be a constant

runtime-util/src/main/java/io/camunda/connector/runtime/util/outbound/ConnectorResult.java
```diff
@@ -0,0 +1,55 @@
...
+
+  public void setException(Exception exception) {
+    this.exception = exception;
+  }
+
+  public Exception getException() {
+    return exception;
+  }
+
+  public void setResponseValue(Object responseValue) {
```
menski, line 44: nitpick: maybe move setter and getter together

menski (dismissed):
Thanks looks good for me, just added some small comments but nothing which should block a merge.

runtime-util/src/main/java/io/camunda/connector/runtime/util/feel/FeelConnectorFunctionProvider.java
tmetzke: 👍

runtime-util/src/main/java/io/camunda/connector/runtime/util/feel/FeelConnectorFunctionProvider.java
tmetzke: 👍

runtime-util/src/main/java/io/camunda/connector/runtime/util/outbound/ConnectorResult.java
tmetzke: 👍

runtime-util/src/main/java/io/camunda/connector/runtime/util/ConnectorHelper.java
```diff
@@ -52,17 +54,26 @@ public static Map<String, Object> createOutputVariables(
...
   }
 
-  @SuppressWarnings("unchecked")
-  private static Map<String, Object> parseJsonVarsAsMapOrThrow(
-      final String jsonVars, final String expression) {
+  public static Optional<BpmnError> examineErrorExpression(
+      final Object responseContent, final Map<String, String> jobHeaders) {
+    final var errorExpression = jobHeaders.get(ERROR_EXPRESSION_HEADER_NAME);
+    return Optional.ofNullable(errorExpression)
+        .filter(s -> !s.isBlank())
```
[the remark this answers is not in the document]
tmetzke: Definitely, nice catch! 👍

PR #83, opened 2022-09-18
feat: support custom container handler class for @Secret

### LINKED ISSUE #81: Allow custom secret container handling

I want to hold secrets in a container class that is not handled by the SDK as I need it. If I have secrets somewhere in a JSON object, I would like to replace them in the JSON directly.

I can add a handler class to the `@Secret` annotation. This handler allows to define how to traverse the container and extract elements to inspect for secrets. From the handler, I can call a `handleElement` method that takes care of the element as it's done in the standard SDK routine.

* I can leave out the annotation for my object and use the secret store from the context object to do secret replacement manually. I would however like to reuse the existing replacement capabilities.
* I can transform the JSON to a Java Map and annotate that map for holding secrets. This includes unnecessary transformations though. I might have to transform the map back to JSON afterwards, for example.

__Additional context__

This might be needed in the REST Connector.

tmetzke:
## Description

Allows defining a custom container handler for Secret containers in the `@Secret` annotation.

## Related issues

closes #81

### DIFF
--- a/core/src/main/java/io/camunda/connector/impl/ReflectionHelper.java
+++ b/core/src/main/java/io/camunda/connector/impl/ReflectionHelper.java
@@ -1,0 +1,148 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH
+ * under one or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information regarding copyright
+ * ownership. Camunda licenses this file to you under the Apache License,
+ * Version 2.0; you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.camunda.connector.impl;
+
+import java.lang.reflect.Field;
+import java.lang.reflect.Method;
+import java.lang.reflect.Modifier;
+import java.util.Arrays;
+
+/** Utility class for reflection helper methods, like finding getters and setter for fields. */
+public class ReflectionHelper {
+
+  private ReflectionHelper() {}
+
+  /**
+   * Gets the value of a field for an input object. Either, the field 'example' is accessible or the
+   * object provides a getter that is named 'getExample' or 'isExample'.
+   *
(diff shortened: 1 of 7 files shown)


core/src/main/java/io/camunda/connector/impl/ReflectionHelper.java
```diff
@@ -0,0 +1,136 @@
...
+ */
+package io.camunda.connector.impl;
+
+import java.lang.reflect.Field;
+import java.lang.reflect.Method;
+import java.lang.reflect.Modifier;
+import java.util.Arrays;
+
+/** Utility class for reflection helper methods, like finding getters and setter for fields. */
+public class ReflectionHelper {
```
igpetrov, line 25: Consider making it final

```suggestion
public final class ReflectionHelper {
```

igpetrov: Also, this class is quite complex. Could we add some unit tests for it?

tmetzke: It's actually tested quite extensively in the [OutboundConnectorContextTest](https://github.com/camunda/connector-sdk/pull/83/files#diff-263b4930ecc0ffdb06fbacf3f15dd947b3beddd989030662f97b9bfb755af1c0)

core/src/main/java/io/camunda/connector/impl/secret/SecretHandler.java
```diff
@@ -0,0 +1,175 @@
...
+      handleSecretsArray((Object[]) input);
+    } else if (input instanceof Map) {
+      handleSecretsMap((Map<Object, Object>) input);
+    } else if (input instanceof List) {
+      handleSecretsList((List<Object>) input);
+    } else if (input instanceof Iterable) {
+      handleSecretsIterable((Iterable<?>) input);
+    } else {
+      handleSecretsField(input);
+    }
```
igpetrov, line 71: Do we really need to handle separately array, list and iterable? Feels like we could simplify this branching.

core/src/main/java/io/camunda/connector/impl/ReflectionHelper.java
igpetrov: Thanks for confirming.

igpetrov (approved):
LGTM

igpetrov (approved):
Will this also solve https://github.com/camunda/connector-http-json/issues/16?

tmetzke:
> Will this also solve [camunda/connector-http-json#16](https://github.com/camunda/connector-http-json/issues/16)?

Not per se, I think. But it will make it easier to handle. You can for example provide a custom secret handling for JSON content now. If there is enough demand, we could also provide it from the SDK as a convenience.

PR #99, opened 2022-09-20
chore: load EL provider dynamically, exclude by default

### LINKED ISSUE #98: Remove dependency on EL provider in validation
__What should we do?__

- Remove the dependency on `org.glassfish:jakarta.el` in the `connector-validation`
- Allow providing an EL provider from a Connector if needed
- Dynamically load the EL provider and support expressions in validation error messages

__Why should we do it?__

Jakarta EL comes with EPL 2.0 which we usually try to avoid re-distributing. Plus, we don't need expressions by default in validation messages. It's convenient enough to add an EL provider if needed.

tmetzke:
## Description

Remove EL in validation by default and load dynamically if provided.

## Related issues

closes #98

### DIFF
--- a/parent/pom.xml
+++ b/parent/pom.xml
@@ -75,7 +75,6 @@ limitations under the License.</license.inlineheader>
     <version.assertj>3.23.1</version.assertj>
     <version.jackson-module-scala>2.13.4</version.jackson-module-scala>
     <version.hibernate-validator>8.0.0.Final</version.hibernate-validator>
-    <version.glassfish-el>4.0.2</version.glassfish-el>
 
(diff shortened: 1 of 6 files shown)


tmetzke:
I did not include an automated test that checks if expressions work when an EL provider is present. I manually tested this with the Connector template by adding `org.glassfish.expressly:expressly` and `org.glassfish:jakarta-el`. Both providers worked fine.

jonathanlukas (approved):
Looks like it wasn't too much effort to exclude EL from Validation after all..

Sorry I did not make this step when I implemented it.

PR #109, opened 2022-09-21
chore: structure in `impl` + `api`

nikku:
## Description

Proposal to re-structure the job worker run-time in accordance with our `impl` / `api` package pattern.

### DIFF
--- a/runtime-job-worker/src/main/java/io/camunda/connector/runtime/jobworker/api/outbound/ConnectorJobHandler.java
+++ b/runtime-job-worker/src/main/java/io/camunda/connector/runtime/jobworker/api/outbound/ConnectorJobHandler.java
@@ -15,7 +15,7 @@
  * limitations under the License.
  */
 
-package io.camunda.connector.runtime.jobworker.outbound;
+package io.camunda.connector.runtime.jobworker.api.outbound;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
@@ -23,8 +23,8 @@ import io.camunda.connector.api.outbound.OutboundConnectorFunction;
 import io.camunda.connector.api.secret.SecretProvider;
 import io.camunda.connector.api.secret.SecretStore;
 import io.camunda.connector.impl.outbound.AbstractOutboundConnectorContext;
-import io.camunda.connector.runtime.jobworker.feel.FeelEngineWrapper;
-import io.camunda.connector.runtime.jobworker.feel.FeelEngineWrapperException;
+import io.camunda.connector.runtime.jobworker.impl.feel.FeelEngineWrapper;
+import io.camunda.connector.runtime.jobworker.impl.feel.FeelEngineWrapperException;
 import io.camunda.zeebe.client.api.response.ActivatedJob;
 import io.camunda.zeebe.client.api.worker.JobClient;
 import io.camunda.zeebe.client.api.worker.JobHandler;
--- a/runtime-job-worker/src/main/java/io/camunda/connector/runtime/jobworker/Main.java
+++ b/runtime-job-worker/src/main/java/io/camunda/connector/runtime/jobworker/Main.java
@@ -18,8 +18,8 @@
 package io.camunda.connector.runtime.jobworker;
 
 import io.camunda.connector.api.outbound.OutboundConnectorFunction;
-import io.camunda.connector.runtime.jobworker.outbound.ConnectorJobHandler;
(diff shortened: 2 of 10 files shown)


runtime-job-worker/src/main/java/io/camunda/connector/runtime/jobworker/impl/outbound/feel/FeelEngineWrapper.java
```diff
@@ -14,7 +14,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package io.camunda.connector.runtime.jobworker.feel;
+package io.camunda.connector.runtime.jobworker.impl.outbound.feel;
```
tmetzke, line 17: Looking at how Bernd worked with this, we will probably reuse the wrapper also for inbound. I think we can actually keep it without `outbound`. But in the end, it's not API, so wet could also change this later as well.

runtime-job-worker/src/main/java/io/camunda/connector/runtime/jobworker/impl/outbound/feel/FeelEngineWrapperException.java
```diff
@@ -14,7 +14,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package io.camunda.connector.runtime.jobworker.feel;
+package io.camunda.connector.runtime.jobworker.impl.outbound.feel;
```
tmetzke, line 18: Same as the wrapper class.

runtime-job-worker/src/main/java/io/camunda/connector/runtime/jobworker/api/outbound/ConnectorJobHandler.java
```diff
@@ -15,16 +15,16 @@
  * limitations under the License.
  */
 
-package io.camunda.connector.runtime.jobworker.outbound;
+package io.camunda.connector.runtime.jobworker.api.outbound;
```
tmetzke, line 18: We need to move the class accordingly as well then.

tmetzke (commented):
I like it, way more consistent 👍
The FEEL classes don't have to move necessarily, I think. And one of the classes needs to be moved according to the new package, if I'm not mistaken.

runtime-job-worker/src/main/java/io/camunda/connector/runtime/jobworker/api/outbound/ConnectorJobHandler.java
nikku: I'm honestly not sure what IDEA is doing here. Why do I use an IDE in the first place if I cannot rely on it to "MOVE" things properly. :see_no_evil:

runtime-job-worker/src/main/java/io/camunda/connector/runtime/jobworker/impl/outbound/feel/FeelEngineWrapper.java
nikku: Agreed + we can likely make it a utility in connectors generally at some point.

This is missing in the Camunda stack: A proper Java wrapper around feel-scala.

runtime-job-worker/src/main/java/io/camunda/connector/runtime/jobworker/api/outbound/ConnectorJobHandler.java
nikku: Class moved.

runtime-job-worker/src/main/java/io/camunda/connector/runtime/jobworker/impl/outbound/feel/FeelEngineWrapper.java
nikku: Moved back out of outbound. We can move these folks around anytime in the future again.

runtime-job-worker/src/main/java/io/camunda/connector/runtime/jobworker/impl/outbound/feel/FeelEngineWrapperException.java
nikku: Moved.

tmetzke (approved):
Nice 👍

tmetzke:
@nikku, can you add this to the update guide in https://github.com/camunda/camunda-platform-docs/pull/1255 as well?

nikku:
Done via https://github.com/camunda/camunda-platform-docs/pull/1255/commits/3b0763da438c8c86163bff75827fa38871453077, please have a look.

Issue #12, opened 2022-07-08
Be able to provide connector default meta-data


nikku:
__What should we do?__

Be able to provide connector meta-data in a structured manner so that run-times can consume it (to load fetched variables / name / type).

__Describe the solution you'd like__

This could work by annotating a connector function to define meta data:

```java
@OutboundConnector(
        name = "amqp",
        taskType = "io.camunda.community:amqp:1",
        inputVariables = {"destinationName","destinationType","username","password"})
class MyConnectorFunction implements ConnectorFunction {
    ...
}
```

* [x] add `OutboundConnector` annotation https://github.com/camunda/connector-sdk/pull/104
* [ ] pick up `OutboundConnector` annotation meta-data in jobworker run-time https://github.com/camunda/connector-sdk/pull/111

__Why should we do it?__

Be able to parse default configuration from connector functions so they can be consumed safely , i.e. in a run-time that batch-executes a number of them.

nikku:
[An instance](https://camunda.slack.com/archives/C02JLRNQQ05/p1660954753577799?thread_ts=1660754058.854809&cid=C02JLRNQQ05) where a user is struggling with this.

nikku:
@igpetrov @tmetzke Right now we parse outbound connectors (to be registered run with job-worker built-in runtime) from environment variables.

The user is in full control of which connector to run and how to configure it (https://github.com/camunda/connector-sdk/tree/main/runtime-job-worker#running-connector-functions). You could argue that _at least_ `inputVariables` are hard-coded in the connector (shall always be fetched from the annotation if present). What about the other ones?

* Shall we auto-discover connectors on the classpath (+ configure them automatically?). We could allow users to opt-out but auto-discovery (via SPI) would be the default
* _OR_ Shall the user still explicitly set which connectors to execute, i.e. via a `ZEEBE_CONNECTORS=io.camunda.Foo:1,io.camunda.Echo:10` environment variable => Backwards compatible + reads existing environment variables, too, if present, to override provided defaults

To aid migration I'd print a warning if the job worker does not define meta-data via `@OutboundConnector` annotation.

igpetrov:
Thank you, @nikku for putting it all together. I am a bit of out of the context but about how job worker actually works.
Assuming, Job Worker can contain multiple `@OutboundConnector`s and the goal it to target self-managed platforms.

[...]

PR #111, opened 2022-09-22
Automatic discovery of Connectors in Job Worker Run-Time based

nikku:
## Description

Adds auto-discovery of connectors to the job worker run-time:

* instantiates all `OutboundConnectorFunction` SPI providers
* picks up our new `@OutboundConnector` annotation (https://github.com/camunda/connector-sdk/issues/12#issuecomment-1254830619).
* manual configuration mode remains un-touched (we'll either use automatic _OR_ manual mode)
* changed prefix for connector related ENV variables from `ZEEBE_` to `CONNECTOR_`

## Related issues

Closes https://github.com/camunda/connector-sdk/issues/12


nikku:
From what it looks like I'd need to bootstrap a full fledged integration test to verify the behavior (cannot mock System#getEnv): https://github.com/camunda/connector-sdk/compare/connector-auto-discovery-test?expand=1

runtime-job-worker/src/main/java/io/camunda/connector/runtime/jobworker/Main.java
```diff
@@ -66,23 +64,21 @@ public static void main(String[] args) {
...
+                  registration -> {
                     LOGGER.info(
-                        "Registering connector function {} as {} on job type {} with variables {}",
-                        connector.getFunction(),
-                        connector.getName(),
-                        connector.getType(),
-                        connector.getVariables());
-
-                    var connectorFunction = loadConnectorFunction(connector.getFunction());
+                        "Registering registration function {} as {} on job type {} with variables {}",
```
tmetzke, line 69:
```suggestion
                        "Registering Connector function {} as {} on job type {} with variables {}",
```

runtime-job-worker/src/main/java/io/camunda/connector/runtime/jobworker/impl/outbound/OutboundConnectorRegistration.java
```diff
@@ -0,0 +1,221 @@
...
+        | InstantiationException
+        | IllegalAccessException
+        | ClassCastException
+        | NoSuchMethodException e) {
+      throw loadFailed("Failed to load " + clsName, e);
+    }
+  }
+
+  private static RuntimeException loadFailed(String s, Exception e) {
+    throw new IllegalStateException(s, e);
```
tmetzke, line 184:
```suggestion
    return new IllegalStateException(s, e);
```

tmetzke (approved):
This looks good to me so far 👍 
Regarding the ENV vars, we could use [System Stubs](https://www.baeldung.com/java-system-stubs), for example. I won't be able to look into this until later tonight. @nikku, if you have time and want to go ahead, feel free. Otherwise, I'll look into it this evening (CEST).

nikku:
Pushed test coverage on top (https://github.com/camunda/connector-sdk/pull/111/commits/e6506b4d1bde39713ceda29ad7c52cd2036d04f0). Not sure how that magic utility does it, but it seems to get the trick done.

runtime-job-worker/pom.xml
```diff
@@ -51,6 +55,12 @@
       <scope>test</scope>
     </dependency>
 
+    <dependency>
+      <groupId>uk.org.webcompere</groupId>
+      <artifactId>system-stubs-core</artifactId>
+      <version>${system-stubs.version}</version>
```
tmetzke, line 61: Let's add test scope here.
```suggestion
      <version>${system-stubs.version}</version>
      <scope>test</test>
```

nikku: Cf. https://github.com/camunda/connector-sdk/pull/111/commits/30dbc88ae1a50b49783d9f444bde86ce7d0fb5ea.

runtime-job-worker/src/main/java/io/camunda/connector/runtime/jobworker/impl/outbound/OutboundConnectorRegistration.java
nikku: Cf. https://github.com/camunda/connector-sdk/pull/111/commits/30dbc88ae1a50b49783d9f444bde86ce7d0fb5ea.

nikku:
Squashed / cleaned up via https://github.com/camunda/connector-sdk/pull/111/commits/d03baa693e3b869e75438fcea1fae3539a386532.

Issue #112, opened 2022-09-22
Add the ability to ignore validate and replaceSecrets Methods for test


Oleksiivanov:
_What should we do_
Add the ability to create connector context with the ability to ignore validate and replaceSecrets Methods for the test.

_Why we do it_

I think it would be better if we extend the `contextBuilder` in the SDK to disable validation and secret replacement for such scenarios. It's a common use case, I think. So our test tool should support this. Can you create an issue for this? If you like, feel free to also propose how to support this in the `contextBuilder`.

_Originally posted by @tmetzke in https://github.com/camunda/connector-sendgrid/pull/100#discussion_r976848414_

tmetzke:
Closing this as "Won't do", see also [PR comment](https://github.com/camunda/connector-sdk/pull/113#issuecomment-1257497815).

PR #113, opened 2022-09-22
feat(test context): add ability ignore validate and replaceSecrets

Oleksiivanov:
## Description
Add the ability to create connector context with the ability to ignore validate and replaceSecrets Methods for the test.

## Related issues

closes #
https://github.com/camunda/connector-sdk/issues/112


nikku:
I do understand where we'Re coming from in https://github.com/camunda/connector-sendgrid/pull/100/files#diff-bb968d3bc500d11c419a139101e341c59f8b912e7a08fb58dd930f50853cc426.

What I see is that we want to test behavior that is independent of the actual request, right?

I still believe it is better to feed a valid input into such methods. Is there a reason we should not? We generally try to test real things ("integration test") over extensive mocking in unit tests.

Hence I'd not add such test utility.

nikku:
I'm simply consulting here and I'd love you and @tmetzke as well as @igpetrov to carefully decide on the direction you want to go.

My main point (of rather general nature) is: Let's aim for integration over unit tests; reduce ad hoc mocking and test-only code paths to a minimum, test scenarios using real-world inputs.

tmetzke:
Thanks for the input, @nikku, I think that is a valid point. After your comment, I looked into the issue a bit more and came to agree with what you recommended.

@Oleksiivanov, let's create a follow-up for the SendGrid Connector tests. They should only use valid input to test the Connector function. It makes more sense to test the behavior of the Connector function with input that it can actually receive and that will be processed after successful validation. Let's not mock our way around such situations.

Closing this here as "Won't do".

PR #114, opened 2022-09-26
Provide jar with dependencies for job worker runtime

menski:
## Description

- provides a jar with dependencies for job worker runtime needed for shipping connectors On-Premises
- upload all jar files from the target folder to the github release
- adds an env variable `ZEEBE_INSECURE` to connect to a local installation using plain text

## Related issues

related to https://github.com/camunda/team-connectors/issues/99

### DIFF
--- a/.github/workflows/RELEASE.yml
+++ b/.github/workflows/RELEASE.yml
@@ -161,9 +161,6 @@ jobs:
           tag_name: ${{ github.event.inputs.version }}
           target_commitish: ${{ needs.prepare.outputs.branch }}
           files: |
-            **/runtime-job-worker/target/*[0-9].jar
-            **/runtime-job-worker/target/*-rc[0-9].jar
-            **/runtime-job-worker/target/*-alpha[0-9].jar
             **/THIRD_PARTY_NOTICES
 
   next-version:
--- a/runtime-job-worker/pom.xml
+++ b/runtime-job-worker/pom.xml
@@ -64,4 +64,12 @@
     </dependency>
   </dependencies>
 
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-shade-plugin</artifactId>
+      </plugin>
+    </plugins>
+  </build>
 </project>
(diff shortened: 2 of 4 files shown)


runtime-job-worker/pom.xml
```diff
@@ -64,4 +64,55 @@
...
+                <transformer implementation="org.apache.maven.plugins.shade.resource.ApacheNoticeResourceTransformer">
+                  <addHeader>false</addHeader>
+                </transformer>
+              </transformers>
+            </configuration>
+          </execution>
+        </executions>
+      </plugin>
+    </plugins>
+  </build>
```
tmetzke, line 116: Should we move this into the `parent` as a plugin management and then just call from here? Like that, we can reuse that in the OOTB Connectors directly, I believe.

tmetzke: We could place this behind a profile like `fat-jar`, so we only execute this when we really need this. We could also leave the standard artifact name in that case maybe. In the release process, we can first build the whole thing without profile and deploy to Maven Central (so you can pick it up as a dependency as a thin JAR). Afterward, we build with the `fat-jar` profile, override the artifact, and build a GitHub release with that. WDYT?

.github/workflows/RELEASE.yml
```diff
@@ -161,9 +161,7 @@ jobs:
           tag_name: ${{ github.event.inputs.version }}
           target_commitish: ${{ needs.prepare.outputs.branch }}
           files: |
-            **/runtime-job-worker/target/*[0-9].jar
-            **/runtime-job-worker/target/*-rc[0-9].jar
-            **/runtime-job-worker/target/*-alpha[0-9].jar
+            **/runtime-job-worker/target/*.jar
```
tmetzke, line 164: This will also attach sources and Javadoc JARs to the release. I would actually like to avoid that. Have a look at my proposal regarding the shade plugin, this might solves this here.

runtime-job-worker/pom.xml
menski: I pushed a new commit with our agreed approach:
- no profile needed
- parent pom contains configuration
- runtime only calls plugin during build
- no upload for jars to github releases, instead people can download from maven central

.github/workflows/RELEASE.yml
menski: We decided to not attach jars into the github release, instead people can download the artifacts from maven central

menski:
@tmetzke I pushed an update based on our conversation

tmetzke (approved):
:shipit:

PR #115, opened 2022-09-26
Add dockerfile for jobworker runtime distro

menski:
## Description

Add Dockerfile for jobworker runtime including out of the box connectors.

## Related issues

related to https://github.com/camunda/team-connectors/issues/99


menski:
Note I will describe in a follow up PR for the OOTB connector bundle how to use this image.

tmetzke (approved):
Looks nice! 👍 
Is there anything we need to do to push the `latest` tag as well to the latest release? Or is this done somewhere else? Or will we not have a `latest` tag?

menski:
Right now it doesn't push a latest tag, I wasn't sure but I guess from your question it seems to be expected.

tmetzke:
We can also not provide it right now and wait for people to ask for it. I'm fine with that.

menski:
I added the latest tag, as I think it is common

.github/workflows/RELEASE.yml
```diff
@@ -139,6 +141,22 @@ jobs:
...
+        with:
+          username: ${{ steps.secrets.outputs.DOCKERHUB_USER }}
+          password: ${{ steps.secrets.outputs.DOCKERHUB_PASSWORD }}
+
+      - name: Build and Push Docker Image
+        uses: docker/build-push-action@v3
+        with:
+          context: runtime-job-worker/
+          push: true
+          tags: camunda/connectors:${{ github.event.inputs.version }} , camunda/connectors:latest
```
tmetzke, line 158: I guess we should only push the `latest` tag when we release from `main`? Currently, we push this tag whenever we build any release, including patches, don't we?

tmetzke: As I said, let's maybe not provide a latest tag yet and see if there is demand for this?

menski: Yes this would push latest on every release. Right now I assume that me only move forward, i.e. our latest release will always be the latest patch or minor. I assume we will not yet start to provide patch releases for previous minor releases, i.e. providing patch 0.1.1 after we release 0.2.0. And my assumption is right now that this will stay like this for a while, i.e. until we hit 1.0.0.

If my assumption is wrong then I would not push latest like this and remove it for now. If my assumption is right then I think it's convenient to provide latest like this for now until 1.0.0

tmetzke: Completely agree with every aspect here. My only concern is that we forget about this in the future. Can we create a follow-up task to tackle this once we need it? With that, I am totally fine to leave it like this.

menski: I created https://github.com/camunda/connector-sdk/issues/127 as follow up issue

PR #117, opened 2022-09-26
chore: account for static fields and methods in secrets

### LINKED ISSUE #116: Account for static members in secret replacement
__What should we do?__

Account for static members in classes when replacing secrets.

Our current take on this:

* You can have static fields in your class.
* You can have static methods in your class.
* We allow retrieving values via static methods.
* We do not allow replacing secrets on static fields.

__Why should we do it?__

Java classes can contain dynamically (but also manually by the user) created `static` members (fields or methods). Those cannot be probed using `canAccess` with an actual non-null input object. They expect to be probed with a `null` object.

tmetzke:
## Description

Handles static members according to #116 

## Related issues

closes #116


nikku (approved):
Works as expected. Thorough test coverage, too :tada:

jonathanlukas (commented):
Very nice catch.

I did not think about POJO classes having static fields (as it makes no sense to me 😓 ), but you never know who will use the API.

PR #122, opened 2022-09-27
Instantiate FEEL engine wrapper statically

### LINKED ISSUE #121: Allow to reuse the Feel Engine between Connector Job Handlers

In the current approach each `ConnectorJobHandler` creates a new [`FeelEngineWrapper`](https://github.com/camunda/connector-sdk/blob/36b8106c6aead76936a00bd4f1be01e9bc44e17e/runtime-job-worker/src/main/java/io/camunda/connector/runtime/jobworker/api/outbound/ConnectorJobHandler.java#L60) which creates a [new FEEL engine](https://github.com/camunda/connector-sdk/blob/36b8106c6aead76936a00bd4f1be01e9bc44e17e/runtime-job-worker/src/main/java/io/camunda/connector/runtime/jobworker/impl/feel/FeelEngineWrapper.java#L40-L43) internally. This means for each connector registered in the runtime one FEEL engine is created. In our OOTB connector bundle for the getting started experience this corresponds to 6 FEEL engines. It is not needed that every connector has their own FEEL engine, and we could reduce the resources used by reusing the FEEL engine. 

I want to be able to ensure that a single FEEL engine is created in my connector bundled runtime and shared by all connectors.

- Keep it as it is and create a new engine for each connector
- Add FEEL engine wrapper to public API signature to allow to inject a reused wrapper, disadvantage exposing of internal classes in a public API, see https://github.com/camunda/connector-sdk/pull/118
- Make FEEL engine wrapper a singleton to be used

__Additional context__

- [Reverted PR](https://github.com/camunda/connector-sdk/pull/118)
- [Slack Conversation](https://camunda.slack.com/archives/C0350959WG1/p1664222461893879)

menski:
## Description

To reduce the number of FEEL engines created we can instantiate the FEEL engine wrapper statically in the ConnectorJobHandler similar to the ObjectMapper. The FEEL engine wrapper bundles a FEEL engine and an ObjectMaper both of them should be thread safe.

## Related issues

closes #121

### DIFF
--- a/runtime-job-worker/src/main/java/io/camunda/connector/runtime/jobworker/api/outbound/ConnectorJobHandler.java
+++ b/runtime-job-worker/src/main/java/io/camunda/connector/runtime/jobworker/api/outbound/ConnectorJobHandler.java
@@ -45,10 +45,10 @@ public class ConnectorJobHandler implements JobHandler {
   protected static final String RESULT_VARIABLE_HEADER_NAME = "resultVariable";
   protected static final String RESULT_EXPRESSION_HEADER_NAME = "resultExpression";
 
+  private static final FeelEngineWrapper FEEL_ENGINE_WRAPPER = new FeelEngineWrapper();
   private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
 
   private final OutboundConnectorFunction call;
-  private final FeelEngineWrapper feelEngineWrapper;
 
   /**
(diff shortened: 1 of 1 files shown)


tmetzke:
I might be overlooking something here, but don't we create one wrapper per handler still like this?

nikku (approved):

nikku (approved):

PR #126, opened 2022-09-27
Provide OOTB connector docker image

### LINKED ISSUE #124: Provide docker image bundling all OOTB connectors

As a user I want to be able to easily try out the Camunda provided out of the box connectors in my local (docker compose) or self-managed setup (helm k8s). Therefore we want to provide a docker image including all OOTB connectors on docker hub.

Provide a OOTB connectors docker image to include the job worker runtime and all OOTB connectors. As a user I can use this image to try out the connectors. I can use environment variables to inject secrets into the connector runtime. I can access this image on docker hub under `camunda/connectors`. The image will be tagged as `ootb` flavor with `SNAPSHOT-ootb` for every push to the `main` branch and on release a new image with the release version is pushed to docker hub.

Additionally to the OOTB connectors image we will provide a description in the README how to extend the base image to include custom connectors.
 
Customers have to always extend the base image and either build a downstream image including connectors or linking them using volumes.

menski:
## Description

- Add image to bundle all OOTB connectors
- Build image for SNAPSHOTS and releases
- Describe how to extend base image and use OOTB image including secrets

## Related issues

closes #124 
based on #115

### DIFF
--- a/.github/workflows/RELEASE_OOTB_IMAGE.yaml
+++ b/.github/workflows/RELEASE_OOTB_IMAGE.yaml
@@ -1,0 +1,66 @@
+name: Build and Push OOTB Connectors Docker Image
+
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: 'Semantic version number of job worker runtime to use in release: ^[0-9]+.[0-9]+.[0-9]+(-[a-zA-Z0-9.-]+){0,1}$'
+        required: true
+      branch:
+        description: 'Git branch of Dockerfile to be used in the release'
+        default: 'main'
+      push-latest:
+        description: 'Push docker image with latest tag, i.e. this is the latest stable version released'
+        type: boolean
+        default: true
+
+jobs:
+  build-and-push:
+    name: Build and Push OOTB Connectors Docker Image
+    runs-on: ubuntu-latest
+    needs: [ prepare ]
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          ref: ${{ github.event.inputs.branch }}
+
+      - name: Import Secrets
+        id: secrets
+        uses: hashicorp/vault-action@v2.4.2
+        with:
+          url: ${{ secrets.VAULT_ADDR }}
+          method: approle
+          roleId: ${{ secrets.VAULT_ROLE_ID }}
+          secretId: ${{ secrets.VAULT_SECRET_ID }}
+          exportEnv: false # we rely on step outputs, no need for environment variables
+          secrets: |
+            secret/data/products/connectors/ci/common DOCKERHUB_USER;
+            secret/data/products/connectors/ci/common DOCKERHUB_PASSWORD;
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+
+      - name: Login to Docker Hub
+        uses: docker/login-action@v2
+        with:
+          username: ${{ steps.secrets.outputs.DOCKERHUB_USER }}
+          password: ${{ steps.secrets.outputs.DOCKERHUB_PASSWORD }}
+
+      - name: Build and Push Docker Image tag ${{ github.event.inputs.version }}-ootb
+        uses: docker/build-push-action@v3
+        with:
+          context: runtime-job-worker/
+          file: runtime-job-worker/Dockerfile.ootb
+          build-args: RUNTIME_VERSION=${{ github.event.inputs.version }}
+          push: true
+          tags: camunda/connectors:${{ github.event.inputs.version }}-ootb
+
+      - name: Build and Push Docker Image with tag latest-ootb
+        if: ${{ github.event.inputs.push-latest }}
+        uses: docker/build-push-action@v3
+        with:
+          context: runtime-job-worker/
+          file: runtime-job-worker/Dockerfile.ootb
+          build-args: RUNTIME_VERSION=${{ github.event.inputs.version }}
+          push: true
+          tags: camunda/connectors:latest-ootb
--- a/runtime-job-worker/README.md
+++ b/runtime-job-worker/README.md
@@ -79,6 +79,42 @@ java -cp 'connector-runtime-job-worker.jar;cloud-connector-slack.jar' \
     io.camunda.connector.runtime.jobworker.Main
 ```
 
+### Docker
+
+#### Docker Job Worker Runtime Image
+
+The [`Dockerfile`](./Dockerfile) in this repository provides a base image
+including the job worker runtime. The image starts the job worker runtime with
+all `jar` files provided in the `/opt/app` directory as classpath.
+
+To use the image at least one connector has to be added to the classpath. We recommend to provide jars with all dependencies bundled.
+
+> :warning: As all connectors share a single classpath it can happen that
+> different versions of the same dependency are available which can lead to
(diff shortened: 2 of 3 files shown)


menski:
Please not this PR currently uses the `rc1` versions of the connectors, and the misaligned group ids, i.e. `io.camunda` vs `io.camunda.connector` once we release the final artifacts we have to align this.

runtime-job-worker/README.md
```diff
@@ -79,6 +79,42 @@ java -cp 'connector-runtime-job-worker.jar;cloud-connector-slack.jar' \
...
+#### Docker Job Worker Runtime Image
+
+The [`Dockerfile`](./Dockerfile) in this repository provides a base image
+including the job worker runtime. The image starts the job worker runtime with
+all `jar` files provided in the `/opt/app` directory as classpath.
+
+To use the image at least one connector has to be added to the classpath. We recommend to provide jars with all dependencies bundled.
+
+> :warning: As all connectors share a single classpath it can happen that
+> different version of the same dependency are available which can lead to
```
tmetzke, line 93:
```suggestion
> different versions of the same dependency are available which can lead to
```

runtime-job-worker/README.md
```diff
@@ -79,6 +79,42 @@ java -cp 'connector-runtime-job-worker.jar;cloud-connector-slack.jar' \
...
+> conflicts. To prevent this common dependency like `jackson` can be shaded and
```
tmetzke, line 94:
```suggestion
> conflicts. To prevent this, common dependencies like `jackson` can be shaded and
```

runtime-job-worker/README.md
```diff
@@ -79,6 +79,42 @@ java -cp 'connector-runtime-job-worker.jar;cloud-connector-slack.jar' \
...
+
+Example adding a connector jar by using volumes
+
+```bash
+docker run --rm --name=connectors -d -v $PWD/connector.jar:/opt/app/ camunda/connectors:0.2.0
+```
+
+#### Docker OOTB Connector Image
+
+The [`Dockerfile.ootb`](./Dockerfile.ootb) in this repository provides a image including the job worker runtime and all out of the box connectors provided by Camunda.
```
tmetzke, line 113:
```suggestion
The [`Dockerfile.ootb`](./Dockerfile.ootb) provides an image including the job worker runtime and all out-of-the-box Connectors provided by Camunda.
```

runtime-job-worker/README.md
```diff
@@ -89,6 +125,22 @@ export MY_SECRET='foo'
 
 Reference the secret in the request payload prefixed with `secrets.MY_SECRET`.
 
+### Docker Image Secrets
+
+To inject secrets into the [docker images of the runtime](#docker) they have to be available in the enviornment of the docker container.
```
tmetzke, line 130:
```suggestion
To inject secrets into the [docker images of the runtime](#docker), they have to be available in the environment of the docker container.
```

runtime-job-worker/README.md
```diff
@@ -89,6 +125,22 @@ export MY_SECRET='foo'
...
+
+Example inject secrets when running a container
```
tmetzke, line 132:
```suggestion
For example, you can inject secrets when running a container:
```

runtime-job-worker/README.md
```diff
@@ -89,6 +125,22 @@ export MY_SECRET='foo'
...
+
+```bash
+docker run --rm --name=connectors -d -e MY_SECRET=secret -e SECRET_FROM_SHELL --env-file secrets.txt camunda/connectors:0.2.0-ootb
+```
+
+The secret `MY_SECRET` value is specified directly in the `docker run` call,
+whereas the `SECRET_FROM_SHELL` is injected based on the value in the
+current shell environment when `docker run` is executed. The `--env-file`
+options allows to use a single file with the format `NAME=VALUE` per line
```
tmetzke, line 141:
```suggestion
option allows using a single file with the format `NAME=VALUE` per line
```

.github/workflows/RELEASE.yml
```diff
@@ -150,13 +150,22 @@ jobs:
...
 
+      - name: Build and Push OOTB Docker Image
+        uses: docker/build-push-action@v3
+        with:
+          context: runtime-job-worker/
+          file: runtime-job-worker/Dockerfile.ootb
+          build-args: RUNTIME_VERSION=${{ github.event.inputs.version }}
+          push: true
+          tags: camunda/connectors:${{ github.event.inputs.version }}-ootb
+
```
tmetzke, line 168: One issue I see coming up here is the release procedure: We usually release the SDK first, then update and release the new versions of the OOTBs, and only then can use them here in the Dockerfile. This will be an issue, I suppose?

menski:
@tmetzke I pushed an updated version with your review hints and a new workflow to manually build and push the image. I also updated the artifact URLs and connector versions in the OOTB Dockerfile to match the final release coordinates. I create the follow up image https://github.com/camunda/connector-sdk/issues/128 summarizing our discussion from slack.

.github/workflows/RELEASE_OOTB_IMAGE.yaml
```diff
@@ -0,0 +1,62 @@
...
+          push: true
+          tags: camunda/connectors:${{ github.event.inputs.version }}
+
+      - name: Build and Push Docker Image with tag latest-ootb
+        if: ${{ github.event.inputs.push-latest }}
+        uses: docker/build-push-action@v3
+        with:
+          context: runtime-job-worker/
+          push: true
+          tags: camunda/connectors:latest-ootb
```
tmetzke, line 62: I believe we need to reference the OOTB Dockerfile here in both steps and also push to `camunda/connectors:${{ github.event.inputs.version }}-ootb` instead in the first step?

tmetzke (commented):
As far as I can see, this looks really good. Thanks for putting this together! 🚀 
I have marked one potential issue in the OOTB release action.

.github/workflows/RELEASE_OOTB_IMAGE.yaml
menski: ah yes sorry missed that, also the tag of the first image is wrong :facepalm:

menski: I fixed it, I also forgot to set the build-args for the release version.

.github/workflows/RELEASE_OOTB_IMAGE.yaml
```diff
@@ -0,0 +1,66 @@
...
+          password: ${{ steps.secrets.outputs.DOCKERHUB_PASSWORD }}
+
+      - name: Build and Push Docker Image tag ${{ github.event.inputs.version }}-ootb
+        uses: docker/build-push-action@v3
+        with:
+          context: runtime-job-worker/
+          file: runtime-job-worker/Dockerfile.ootb
+          build-args: RUNTIME_VERSION=${{ github.event.inputs.version }}
+          push: true
+          tags: camunda/connectors:${{ github.event.inputs.version }}
```
tmetzke, line 56: Do we need to tag it as OOTB? We'll override the base image otherwise, right?
```suggestion
          tags: camunda/connectors:${{ github.event.inputs.version }}-ootb
```

tmetzke (approved):
Looks good to me now 👍 
Let's see how this holds up in release testing 🙂

PR #132, opened 2022-09-28
chore: use `deps` type for dependency updates

nikku:
Attemts to address https://github.com/camunda/team-connectors/issues/130. 

Needs in the wild testing.

----

Cf. https://docs.renovatebot.com/configuration-options/#semanticcommitscope


tmetzke:
These parts currently define how commits are mapped to the changelog:
* Commit parser: https://github.com/conventional-commits/parser
* GitHub action: https://github.com/requarks/changelog-action
* Types/headings defined by the action: https://raw.githubusercontent.com/requarks/changelog-action/main/dist/index.js (search for `const types`)

We need to see if this covers it like we would want it to look in the end.

Wild alternative: let's not create a changelog based on PRs/commits but rather based on issues and labels. This is easier to change after the fact (labels can be easily adjusted compared to commit messages). But it requires having issues for all the things you build in a project (which makes sense). Plus, we'd need a solid approach for how we label issues.

tmetzke:
I looked into https://github.com/requarks/changelog-action/blob/main/index.js again. There are quite some specifics in there. 

For example, it only matches commits to PRs if the PR number is named in the commit message in round brackets at the end (i.e. how it's done but default for squash-merging via GitHub PR UI).

Also, if I'm not mistaken, a commit type like `deps` will not end up in the changelog at all. The action is pretty fixed regarding types and headings in the changelog.

That's just two things I found, there probably is more. I did another search for alternative actions that use commit messages for changelogs but didn't find anything else. Most actions generate from issues and their labels.

Regarding the specific change here: I am not sure if this makes things worse (I think dependency updates should occur in the changelog).

Regarding the changelog in general: as stated above, maybe we should rather base it on issues than commits and their messages.

nikku:
Let's close this PR in that case. 

I believe there is value in explicitly marking dependency upgrades, but no need to rush anything here.

I personally like our commit-based changelog generation, if it was comprehensive / good enough. Only looking at issues / PRs may miss certain specifics.

PR #136, opened 2022-09-30
Extracted some logic to own OutboundConnectorFunctionInvoker that can…

berndruecker:
I would like to reuse some code from this runtime around data input/output instead of copying in to spring-zeebe - therefore I need to untangle it from the JobHandler. Until merged, I will copy the code to spring-zeebe - but looking forward to delete it there again.


berndruecker:
@tmetzke I wondered why the connector sdk requires Java 11 - should that not also work with Java 8? I know that this is pretty old - but if there is no harm we made good experiences to keep compiling for also older SDK's in anything that is client-oriented (like java client or spring-zeebe)

tmetzke:
Thanks, @berndruecker! Can you give a little example maybe how and what you would like to reuse? Maybe we can only extract the `createOutputVariables` method to a utility class that holds the `FeelEngineWrapper`? We will probably enrich this soon with `errorExpression` handing then as well, we can refactor the code [in this PR](https://github.com/camunda/connector-sdk/pull/74/files) to go into the utility class then as well. This would also be closer to what we do in the Cloud Bridge, which means we might even consolidate and reuse the SDK code there: https://github.com/camunda/cloud-connector-bridge/pull/108.

Regarding Java 8: we deemed it simply too old now, we will even drop support in C7 soon I think. That says a lot, I believe 😄

berndruecker:
Howdy @tmetzke. 

Multiple threads start to unfold :-) Maybe it make sense to hop on a quick Zoom sync this week?

1. JDK 8 is still supported for the Java client and Spring-Zeebe (see https://docs.camunda.io/docs/reference/supported-environments/#clients). I don't think it is a good idea to drop this support only for the Connector SDK - especially for a good reason. So we should go back to Java 8 support.
2. Here is how spring-zeebe is invoking outbound functions: https://github.com/camunda-community-hub/spring-zeebe/blob/master/client/spring-zeebe/src/main/java/io/camunda/zeebe/spring/client/jobhandling/JobHandlerInvokingSpringBeans.java#L75. Generally I do not care too much how this code looks like exactly, just that there should not be a copy of things (like now: https://github.com/camunda-community-hub/spring-zeebe/tree/master/client/spring-zeebe/src/main/java/io/camunda/zeebe/spring/client/jobhandling/copy
3. I would generally prefer to get rid of the dependency to the JobWorkerHandler anyway - as this for example contains a log4j.xml - which causes trouble in a library (see e.g. https://github.com/camunda-community-hub/spring-zeebe/issues/230#issuecomment-1273266687). 

So maybe for 2 & 3, those helper stuff could end up in the core of the library or some dedicated helper, so I don't have to pull in the job handler runtime for it.

Also it might be worth to discuss, if spring-zeebe could be the default "job-handler-runtime", so I am not sure if we really need a second one?

Best 
Bernd

nikku:
Spring Zeebe is not officially supported by Camunda, right? It is a community project.

berndruecker:
@nikku: Yes, spring-zeebe is officially a community project. **But:**

1. It is the officially recommended client (in get started & co) and thus the default used by our users already today
2. It will be moved to official sooner or later - more a matter of bandwith than anything else
3. I don't see this as any show stopper to use it in the connector runtime, especially looking at version 0.x

Instead of wasting time rewriting what is already there we should invest our efforts in figuring out the important parts for the connectors and build on what is already there - especially if this is what the Java world is using anyway. If we ever encounter the missing "supported component" label for anything with connectors we could still push for it being supported. 

Happy to discuss - but at the moment I don't see any reason for not using spring-zeebe - in contrary, I think it would immediately solve a couple of runtime, config or deployment challenges around e.g. co located connectors.

nikku:
Thanks for clarifying. Under the assumption that it [becomes the official client one day](https://github.com/camunda/connector-sdk/pull/136#issuecomment-1273521093) this makes sense.

Leaving to @tmetzke, @igpetrov and team to further dig into this topic + decide what is reasonable.

berndruecker:
Probably something we should discuss seperately - but with the Spring Boot runtime you could easily build one "Connector Runtime Deployment" by simply adding all connector (functions) you need as dependency - then your Maven build can spill out an uber jar that you can run co-located in your "own datacenter" (typically to have resource access like Kafka endpoints, but also Security Infrastructure like Vault, etc..).

tmetzke:
So, my general view on this:

* Let's still provide the JobHandler wrapper (and related utilities like the FEEL Engine wrapper) in the SDK in a separate module
  * This pulls in dependencies that are not necessary for Connector Function developers, let's not mix up those things
* Let's promote Spring Zeebe as the default Connector runtime if it makes sense (we provide the runtime via Docker and such, this needs to be considered since we still want to provide this in the future)
  * We can get rid of the `Main` method in the SDK module
  * We can get rid of the logging dependency in the SDK module

PR #141, opened 2022-10-04
fix: use colon for classpath resources

menski:
## Description

Fix the classpath examples to use `:` instead of `;`.

## Related issues

related to https://github.com/jwulf/connector-sdk/pull/1


nikku (approved):
Thanks for the fix. Not sure how this slipped my attention...

PR #142, opened 2022-10-04
Add hadolint to lint Dockerfile in CI

menski:
## Description

Add [hadolint](https://github.com/hadolint/hadolint) to CI workflow to ensure we follow best practices for Dockerfiles.


nikku:
Seems like we don't comply (yet).

PR #146, opened 2022-10-04
fix(runtime): use object mapper to convert to map

menski:
## Description

To support POJO responses of the connector function use the jackson object mapper to ensure that response can be casted to a map.

## Related issues

closes #135

Issue #135, opened 2022-09-29
HTTP connector in `camunda/connectors:0.2.0-rc1-ootb` image fails to evaluate response mapping expression


oleschoenburg:
__Describe the Bug__

I'm trying to run the ootb connectors via `camunda/connectors:0.2.0-rc1-ootb`.

Using the rest connector with a result mapping fails:

```
Sep 29, 2022 10:51:16 AM io.camunda.connector.runtime.jobworker.api.outbound.ConnectorJobHandler handle
INFO: Received job 2251799813851702
Sep 29, 2022 10:51:16 AM io.camunda.connector.runtime.jobworker.api.outbound.ConnectorJobHandler handle
SEVERE: Failed to process job 2251799813851702
io.camunda.connector.runtime.jobworker.impl.feel.FeelEngineWrapperException: Failed to evaluate expression '=result.body.id' and context 'HttpJsonResult{status=200, headers={date=Thu, 29 Sep 2022 10:51:16 GMT, transfer-encoding=chunked, keep-alive=timeout=60, content-type=application/vnd.spring-boot.actuator.v3+json, connection=keep-alive}, body={id=310}}', because expression evaluation failed
	at io.camunda.connector.runtime.jobworker.impl.feel.FeelEngineWrapper.evaluateToJson(FeelEngineWrapper.java:84)
	at io.camunda.connector.runtime.jobworker.api.outbound.ConnectorJobHandler.lambda$createOutputVariables$0(ConnectorJobHandler.java:97)
	at java.base/java.util.Optional.map(Unknown Source)
	at io.camunda.connector.runtime.jobworker.api.outbound.ConnectorJobHandler.createOutputVariables(ConnectorJobHandler.java:97)
	at io.camunda.connector.runtime.jobworker.api.outbound.ConnectorJobHandler.handle(ConnectorJobHandler.java:73)
	at io.camunda.zeebe.client.impl.worker.JobRunnableFactory.executeJob(JobRunnableFactory.java:44)
	at io.camunda.zeebe.client.impl.worker.JobRunnableFactory.lambda$create$0(JobRunnableFactory.java:39)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Unknown Source)
	at java.base/java.util.concurrent.FutureTask.run(Unknown Source)
	at java.base/java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(Unknown Source)
[...]

PR #152, opened 2022-10-05
[BACKPORT] Correct result mapping from POJO responses

menski:
## Description

To support POJO responses of the connector function use the jackson object mapper to ensure that response can be casted to a map.

## Related issues

related to #135 
backport of #146


igpetrov (approved):
Thank you, @menski!

Issue #153, opened 2022-10-05
HTTP connector execution fails when response JSON contains `null` values


oleschoenburg:
__Describe the Bug__

Using the http-json connector to query an API that returns JSON where some fields are `null` results in an exception while evaluating an output variable expression that reads like `null` is not supported by the object mapper used to convert the connector result to JSON.

```
io.camunda.connector.runtime.jobworker.impl.feel.FeelEngineWrapperException: Failed to evaluate expression '={status: response.body.status}' and context 'HttpJsonResult{status=200, headers={date=Wed, 05 Oct 2022 13:08:01 GMT, transfer-encoding=chunked, keep-alive=timeout=60, content-type=application/vnd.spring-boot.actuator.v3+json, connection=keep-alive}, body={backupId=700, status=COMPLETED, failureReason=java.lang.Object@157c77c3, partitions=[{partitionId=1, status=COMPLETED, description={snapshotId=839860-2-1159347-1159109, checkpointPosition=1353282, brokerId=1, brokerVersion=8.1.0}, failureReason=java.lang.Object@157c77c3, createdAt=2022-10-04T13:43:52.901068692Z, lastUpdatedAt=2022-10-04T13:43:55.612012619Z}, {partitionId=2, status=COMPLETED, description={snapshotId=958696-1-1321649-1320638, checkpointPosition=1353064, brokerId=1, brokerVersion=8.1.0}, failureReason=java.lang.Object@157c77c3, createdAt=2022-10-04T13:43:52.900913208Z, lastUpdatedAt=2022-10-04T13:43:55.499697877Z}, {partitionId=3, status=COMPLETED, description={snapshotId=887461-1-1231853-1231786, checkpointPosition=1352981, brokerId=2, brokerVersion=8.1.0}, failureReason=java.lang.Object@157c77c3, createdAt=2022-10-04T13:43:52.907161982Z, lastUpdatedAt=2022-10-04T13:43:55.508196412Z}]}}', because expression evaluation failed
	at io.camunda.connector.runtime.jobworker.impl.feel.FeelEngineWrapper.evaluateToJson(FeelEngineWrapper.java:86)
	at io.camunda.connector.runtime.jobworker.api.outbound.ConnectorJobHandler.lambda$createOutputVariables$0(ConnectorJobHandler.java:97)
	at java.base/java.util.Optional.map(Unknown Source)
[...]

PR #148, opened 2022-10-04
chore: gracefully handle all output mapping cases

nikku:
Closes https://github.com/camunda/connector-sdk/issues/145

PR #155, opened 2022-10-10
Ignore unknown object values in response

menski:
## Description

This PR ignores fields of a response, which are not know types for the object mapper. I decided to ignore them to gracefully handle the response instead of failing.

- Add test to verify null values are handled correctly
- Add test to verify unknown object values are ignore

## Related issues

related to #153


tmetzke (approved):
Looks good to me, that should keep objects manageable that we cannot handle automatically 👍

Two notes:
* How is this handled in the SaaS setup? Do we use a similar serialization configuration there?
* Can we align this with #148? I can see some conflicts coming up with those two PRs. Should we backport both? Or merge this one here first and backport it to `0.2`, update #148 afterward and NOT backport #148? @nikku, FYI 🙂

nikku (approved):

nikku:
I#m happy to update #148 afterwards. It mainly streamlines our tests + adds more "expected" coverage. No critical stuff in there (I believe).

menski:
> How is this handled in the SaaS setup? Do we use a similar serialization configuration there?
I actually don't know, never saw this happen before, but I will try to reproduce with the info Ole gave me. The root cause should be the HTTP connector. Will update later today.

tmetzke:
@menski, can you create a backport for `stable/0.2` as well, please? 🙂

PR #156, opened 2022-10-10
[BACKPORT] Add docker image for runtime

menski:
## Description

Backports the docker image from the `main` branch to the `stable/0.2` branch.

## Related issues

backport of #115


.github/workflows/RELEASE.yml
```diff
@@ -139,6 +141,22 @@ jobs:
...
+          username: ${{ steps.secrets.outputs.DOCKERHUB_USER }}
+          password: ${{ steps.secrets.outputs.DOCKERHUB_PASSWORD }}
+
+      - name: Build and Push Docker Image
+        uses: docker/build-push-action@v3
+        with:
+          context: runtime-job-worker/
+          push: true
+          tags: camunda/connectors:${{ github.event.inputs.version }} , camunda/connectors:latest
+
```
tmetzke, line 159: I think it's nice to add this here, but this should actually not be necessary. In fact, we should maybe even delete the release action from maintenance branches. That way, we leave less room for errors for releases (there is always just the `main` release action).

menski: I assumed in the workflow dispatch I can choose the workflow file to use see screenshot? In my experience release processes change over time, and have a single main pipeline build all release branches is making it more complex. 

image

tmetzke: Yes, you can. I agree that this can happen. But this approach then requires that we always update the release action on all relevant branches, right? Like we have to do here now.

menski: yes if we do adjustments to the release pipeline, which are applicable to all releases then we would have to backport that to

tmetzke: OK, just wanted to verify my understanding. Makes sense, I think this is viable. We should document this in the development guidelines, I believe.

tmetzke (approved):

PR #162, opened 2022-10-11
Introduce Connector runtime utility module

berndruecker:
As discussed with @tmetzke we need to refactor utilities required to build runtimes in an own dependency that is not the runtime itself.


pom.xml
```diff
@@ -17,6 +17,7 @@
   <modules>
     <module>parent</module>
     <module>core</module>
+    <module>runtime-util</module>
     <module>runtime-job-worker</module>
```
tmetzke, line 21: If we remove the runtime module for now, we need to remove it here as well.
```suggestion
```

runtime-job-worker/Dockerfile
```diff
@@ -1,7 +0,0 @@
-FROM eclipse-temurin:17.0.4_8-jre
-RUN mkdir /opt/app
-
-COPY target/connector-runtime-job-worker-*-with-dependencies.jar /opt/app
-
-# Using entry point to allow downstream images to add JVM arguments using CMD
-ENTRYPOINT ["java", "-cp", "/opt/app/*", "io.camunda.connector.runtime.jobworker.Main"]
```
tmetzke, line 7: If we remove the runtime and the Dockerfile for now, we need to remove it from the https://github.com/camunda/connector-sdk/blob/main/.github/workflows/DEPLOY.yml and https://github.com/camunda/connector-sdk/blob/main/.github/workflows/RELEASE.yml actions as well.

tmetzke:
One suggestion here to have a step-by-step approach that does not break the SDK along the way:
- Keep the runtime module (and the main) for now
- Introduce the util module
- Reuse the util module in the runtime module

nikku:
@tmetzke I believe this util module could also be off use for us implementing a cloud run-time (and ensuring that both work in the same way)?

tmetzke:
> @tmetzke I believe this util module could also be off use for us implementing a cloud run-time (and ensuring that both work in the same way)?

Yes, exactly, that is a nice side-effect this should have. We should be able to create the util module in a way that the Cloud runtime can consume it as well.

pom.xml
berndruecker: We do not need to remove it right away - I thought we transform it to a Spring-based variant in a second step

runtime-job-worker/Dockerfile
berndruecker: Damn it - I did not want to remove - but move it. Sorry - Git tricked me. PR updated

berndruecker:
@tmetzke 100% That's what I did - unfortunately not exactly what I pushed ;-) PR fixed now - sorry

tmetzke (approved):
Looks good to me 👍 
I can merge this to adjust the commit message to reflect the breaking changes if you like (see our [definition of done](https://github.com/camunda/team-connectors/blob/main/development/DEFINITION_OF_DONE.md)).

berndruecker:
I don't see this as a breaking change - it just introduced one additional modul. Happy to merge it. Thx :-)

tmetzke:
Well, it strictly speaking is not breaking, yeah. It still brings along all the stuff it did before. 👍

PR #171, opened 2022-10-14
Extracted function and renamed packages as discussed

### LINKED ISSUE #164: Refactor `createOutputVariables` into function usable from anywhere else

berndruecker:
Extracted `createOutputVariables` function and refactored packages as discussed in https://github.com/camunda/connector-sdk/pull/166.

closes #164


runtime-util/src/main/java/io/camunda/connector/runtime/util/outbound/OutboundConnectorRegistration.java
```diff
@@ -89,25 +85,27 @@ public static List<OutboundConnectorRegistration> parse() {
...
+                          cfg.getName(), cfg.getType(), cfg.getInputVariables(), function))
+              .orElseThrow(
+                  () ->
+                      new RuntimeException(
+                          String.format(
+                              "OutboundConnectorFunction %s is missing @OutboundConnector annotation",
+                              function.getClass())));
+      foundFunctions.add(outboundConnectorRegistration);
+    }
+    return foundFunctions;
```
tmetzke, line 108: As far as I can see, there is no functional change here, right? Is this left over from the Java 8 restructuring or does this serve a specific purpose? Just want to make sure we don't just accidentally change this.

runtime-util/src/main/java/io/camunda/connector/runtime/util/feel/FeelEngineWrapper.java
```diff
@@ -14,7 +14,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package io.camunda.connector.runtime.jobworker.impl.feel;
+package io.camunda.connector.runtime.util.feel;
```
tmetzke, line 17: All other changes in this class shouldn't be necessary, right?

runtime-util/src/main/java/io/camunda/connector/runtime/util/outbound/JobHandlerContext.java
```diff
@@ -0,0 +1,47 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.camunda.connector.runtime.util.outbound;
+
+import io.camunda.connector.api.secret.SecretStore;
+import io.camunda.connector.impl.outbound.AbstractOutboundConnectorContext;
+import io.camunda.zeebe.client.api.response.ActivatedJob;
+
+public class JobHandlerContext extends AbstractOutboundConnectorContext {
```
tmetzke, line 23: Public class in `util`, let's add some Javadoc here 🙂

runtime-util/src/main/java/io/camunda/connector/runtime/util/outbound/OutboundConnectorRegistration.java
berndruecker: There is no functional change. I kept this code that was indeed already prepped for Java8

runtime-util/src/main/java/io/camunda/connector/runtime/util/outbound/JobHandlerContext.java
berndruecker: Honsetly, if you use this class you should now what it is doing - but let me sse if I can think of anything that adds vaue

runtime-util/src/main/java/io/camunda/connector/runtime/util/feel/FeelEngineWrapper.java
berndruecker: Agreed

berndruecker: Agreed

tmetzke:
Looks good in general. 

Can you have a second look at Java 8-related changes, please? I would like to keep this PR focused on the extraction, everything else can be handled in a Java 8 PR, if we decide that we want to go with this. 

Thank you! I know it's easier to get multiple things out of the way in one PR, but it's simply harder to focus and to ensure we don't mess up anything else 🙂

berndruecker:
Angst? ;-) 
OK - I removed one more J8 change and I think this should be good to go so I would merge it now - OK @tmetzke?

PR #176, opened 2022-10-18
feat(feel): Added evaluations required for inbound connectors

berndruecker:
Moved the code from https://github.com/camunda/connector-sdk-inbound-prototype over here to have everything in one place. 
Missing a test case


runtime-util/src/main/java/io/camunda/connector/runtime/util/feel/FeelEngineWrapper.java
```diff
@@ -40,15 +40,12 @@ public class FeelEngineWrapper {
   private final ObjectMapper objectMapper;
 
   public FeelEngineWrapper() {
-    this(
-        new FeelEngine.Builder()
-            .valueMapper(SpiServiceLoader.loadValueMapper())
-            .functionProvider(SpiServiceLoader.loadFunctionProvider())
-            .build(),
+    this.feelEngine = new FeelEngine.Builder().customValueMapper(new JavaValueMapper()).build();
```
tmetzke, line 43: Could you detail why we want to change this?

berndruecker: Sure: https://github.com/camunda/connector-sdk-inbound-prototype/issues/20#issuecomment-1283406179 (sorry, forget to link it)

tmetzke: OK, I see. This might become obsolete with #74 and the new [`bpmnError` function](https://github.com/camunda/connector-sdk/pull/74/files#diff-a646a7ed4d9b5c162b9da24add41c450f4b0494dcf0cb3d4f7eee2b0b8488c0e) we want to provide.

berndruecker: I don't yet see the connection to the errorFunction - this fixes that we can properly unwrap FEEL evaluation results into Java Maps - which we currently do in Inbound Webhooks.

tmetzke: My bad, mixed the value mapper up with the function mapper. Question now for me is, should I rather hook up the function mapper in #74 statically like we do here with the `JavaValueMapper` or load it via SPI? 🤔

tmetzke: Yeah, definitely. I was just wondering if people would want to provide their own function and value mappers in a custom runtime. That is the only use case that would require this.

tmetzke: Let's revert the constructor then and probably move the custom SPI from `src/test/resources` to `src/main/resources`?

berndruecker: Wow - good catch - fixed location for SPI.

berndruecker: Sorry!

runtime-util/src/main/java/io/camunda/connector/runtime/util/feel/FeelEngineWrapper.java
```diff
@@ -76,13 +77,21 @@ private Map<String, Object> ensureVariablesMap(final Object variables) {
     return objectMapper.convertValue(variables, MAP_TYPE_REFERENCE);
   }
 
+  public <T> T evaluate(final String expression, final Object variables) {
+    return (T) toJava(evaluateInternal(expression, variables));
```
tmetzke, line 81: This should not need conversion to Java, actually, since we already use the `JavaValueMapper`.
```suggestion
    return (T) evaluateInternal(expression, variables);
```

runtime-util/src/main/java/io/camunda/connector/runtime/util/feel/FeelEngineWrapper.java
```diff
@@ -99,4 +108,11 @@ private String resultToJson(final Object result) {
       throw new RuntimeException(e);
     }
   }
+
+  private static Object toJava(final Object scalaObject) {
+    if (scalaObject instanceof scala.collection.immutable.Map) {
+      return CollectionConverters.asJava((scala.collection.immutable.Map) scalaObject);
+    }
+    return scalaObject;
+  }
```
tmetzke, line 117: This should not be necessary when we already use the `JavaValueMapper`. Give it a try, please.
```suggestion
```

runtime-util/src/main/java/io/camunda/connector/runtime/util/feel/FeelEngineWrapperException.java
```diff
@@ -18,6 +18,14 @@
 
 public class FeelEngineWrapperException extends RuntimeException {
 
+  public FeelEngineWrapperException(
+      final String reason, final String expression, final Object context) {
+    super(
+        String.format(
+            "Failed to evaluate expression '%s' and context '%s', because %s",
+            expression, context, reason));
```
tmetzke, line 26: Let's avoid code duplication.
```suggestion
    this(reason, expression, context, null);
```

runtime-util/src/test/java/io/camunda/connector/runtime/util/feel/FeelEngineWrapperExpressionEvaluationTest.java
```diff
@@ -17,6 +17,7 @@
 package io.camunda.connector.runtime.util.feel;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.jupiter.api.Assertions.assertEquals;
```
tmetzke, line 20: Let's stay with one Assertion framework if possible (we could actually get rid of the `assertThrows` as well, but we don't have to do it now.
```suggestion
```

runtime-util/src/test/java/io/camunda/connector/runtime/util/feel/FeelEngineWrapperExpressionEvaluationTest.java
```diff
@@ -49,6 +50,22 @@ void evaluateToJson_ShouldSucceed_WhenHappyCase() {
...
+    final String resultExpression = "{\"processedOutput\": response.callStatus }";
+    // Response from service -> {"callStatus":{"statusCode":"200 OK"}}
+    final Map variables = Map.of("callStatus", Map.of("statusCode", "200 OK"));
+
+    // when
+    final Map evaluatedResultAsMap = objectUnderTest.evaluate(resultExpression, variables);
+
+    // then
+    final Map expectedResult = Map.of("processedOutput", Map.of("statusCode", "200 OK"));
+    assertEquals(expectedResult, evaluatedResultAsMap);
```
tmetzke, line 66: Let's stay with one Assertion framework if possible
```suggestion
    assertThat(evaluatedResultAsMap).isEqualTo(expectedResult);
```

tmetzke (changes requested):
Looks good overall, some suggestions mainly for consistency and code style.

Since we use the `JavaValueMapper` now, we could also expand the test cases a bit to other supported Java types. But this goes a bit into the direction of testing a library we are consuming (FEEL engine library with its `JavaValueMapper`). We might as well rely on the library's tests.

berndruecker:
@tmetzke Changes are accepted and work :-)

tmetzke (approved):
Looks good to me. I agree that this should come with FEEL itself at some point 👍

Issue #178, opened 2022-10-20
Revert FeelEngineWrapper to use hardocded JavaValueMapper


berndruecker:
I vote to change the `FeelEngineWrapper` to use

```java
    feelEngine =
        new FeelEngine.Builder()
            .customValueMapper(new JavaValueMapper())
            .build();
```

as discussed in the recent PR: https://github.com/camunda/connector-sdk/pull/176

With the current version using SPI I have unpredictable behavior depending on which combination of libraries I pull in. 
I don't see any need to customize this at the moment - so I vote to avoid any possible problems down the road.

@tmetzke OK with you? Then I could create a PR for it.

tmetzke:
Hi @berndruecker, this got drowned in my other notifications. 

I am generally fine with adjusting this for now.
However, could you try to reliably reproduce this issue so we can see where this is coming from?
I would love for us to understand why this can become an issue.

berndruecker:
No worries - not urgent. I personally would hope to move all this wrapping to the feel-engine itself one day - so we don't have to think about it outside of it :-)
I haven't followed it so far - it is something with the SPI loading. Don't have enough time to understand the root of it - but will provide a PR for now 👍

PR #184, opened 2022-10-25
fix(feel): Switched to hard coded JavaValueMapper

berndruecker:
fixes https://github.com/camunda/connector-sdk/issues/178


runtime-util/src/main/java/io/camunda/connector/runtime/util/feel/FeelEngineWrapper.java
```diff
@@ -42,9 +43,8 @@ public class FeelEngineWrapper {
   public FeelEngineWrapper() {
     this.feelEngine =
         new FeelEngine.Builder()
-            .valueMapper(SpiServiceLoader.loadValueMapper())
-            .functionProvider(SpiServiceLoader.loadFunctionProvider())
-            .build();
+                .customValueMapper(new JavaValueMapper())
+                .build();
```
tmetzke, line 47:
```suggestion
            .customValueMapper(new JavaValueMapper())
            .build();
```

runtime-util/src/main/java/io/camunda/connector/runtime/util/feel/FeelEngineWrapper.java
```diff
@@ -25,6 +25,7 @@
 import java.util.Map;
 import java.util.Objects;
 import org.camunda.feel.FeelEngine;
+import org.camunda.feel.impl.JavaValueMapper;
 import org.camunda.feel.impl.SpiServiceLoader;
```
tmetzke, line 29:
```suggestion
```

tmetzke:
@berndruecker, you can run `mvn compile` locally (or anything like that includes `compile` like `mvn verify`) to format the code accordingly. 🙂

tmetzke (approved):
Let's hardwire this for now since it seems to make issues. We don't need flexibility here currently.

PR #188, opened 2022-10-25
Working on joined runtime

berndruecker:


inbound/src/main/java/io/camunda/connector/inbound/feel/FeelEngineWrapper.java
```diff
@@ -16,29 +16,24 @@
...
+import org.camunda.feel.FeelEngine;
+import org.camunda.feel.impl.JavaValueMapper;
+import scala.jdk.javaapi.CollectionConverters;
 
 /**
- * Wait for https://github.com/camunda/connector-sdk/issues/178 to be solved - then delete this here.
+ * Wait for https://github.com/camunda/connector-sdk/issues/178 to be solved - then delete this
+ * here.
  */
 public class FeelEngineWrapper {
```
tmetzke, line 31: This can go now, I think 👍

berndruecker: 100%

berndruecker:
Hey @igpetrov I have the big PR ready to be merged. Somehow some wired formatting things were going on when I needed to merge the latest changes into it (my maven build somehow adjusts formatting a little bit different then your code) - I apologize to the additional noise of that. 

Would you be OK to merge this now and take it from there?

runtime-util/src/main/java/io/camunda/connector/runtime/util/outbound/OutboundConnectorRegistrationHelper.java
```diff
@@ -37,8 +37,31 @@ public class OutboundConnectorRegistrationHelper {
...
+      hardwiredEnvironmentVariables = new HashMap<>();
+    }
+    hardwiredEnvironmentVariables.put(key, value);
+  }
+
+  public static void clearHardwiredEnvionmentVariable() {
+    hardwiredEnvironmentVariables = null;
+  }
+
+  public static Map<String, String> getEnvionmentVariables() {
```
tmetzke, line 53: Just a little type there (needs to be fixed in the code using this as well)
```suggestion
  public static Map<String, String> getEnvironmentVariables() {
```

pom.xml
```diff
@@ -17,16 +17,17 @@
   <modules>
     <module>parent</module>
     <module>core</module>
+    <module>inbound</module>
     <module>runtime-util</module>
-    <module>runtime-job-worker</module>
+    <module>runtime</module>
```
tmetzke, line 22: Please keep in mind that this is definitely a breaking change since the artifact `runtime-job-worker` will not exist anymore after this. It's not a big deal, but let's be clear about this in the final commit, please.

runtime/README.md
```diff
@@ -18,8 +18,8 @@ The run-time picks up outbound connectors available on the classpath automatical
 It uses the default configuration specified through the `@OutboundConnector` annotation in these cases.
 
 ```bash
-java -cp 'connector-runtime-job-worker-with-dependencies.jar:connector-http-json-with-dependencies.jar' \
-    io.camunda.connector.runtime.jobworker.Main
+java -cp 'connector-runtime-VERSION.jar:connector-http-json-with-dependencies.jar' \
+    io.camunda.connector.runtime.ConnectorRuntimeApplication
```
tmetzke, line 22: We need to adjust the docs.camunda.io around this as well. Please create a follow-up so we can tackle that. This is currently referenced in https://docs.camunda.io/docs/self-managed/platform-deployment/manual/#run-connectors.

tmetzke (commented):
Looks good from the general runtime perspective (unified runtime, inclusion in the project), I would say 👍 
I haven't looked much at the Inbound stuff, to be honest. @igpetrov, would be nice if you could have a second look at that, thanks 🙂

pom.xml
tmetzke: See also the breaking changes mentioned in https://github.com/camunda/team-connectors/blob/main/development/DEFINITION_OF_DONE.md.

runtime-util/src/main/java/io/camunda/connector/runtime/util/outbound/OutboundConnectorRegistrationHelper.java
berndruecker: Good catch!

pom.xml
berndruecker: Will not be the last breaking change :-) That's the great thing about 0.x anyway. How do you normally deal with those - do you make a list to not forget to announce the right things?

berndruecker: I collect it here: https://github.com/camunda/connector-sdk/issues/203

runtime/README.md
berndruecker: Created https://github.com/camunda/connector-sdk/issues/203

pom.xml
tmetzke: Two ways: 

1. Collect them as you did now if we already know multiple things will come up
2. Handle them right away with the issue if it's one PR/commit that breaks things

In general, we should add the breaking change to the commit, because that is what's ending up in our release notes as well.

inbound/src/main/java/io/camunda/connector/inbound/importer/ProcessDefinitionImporter.java
```diff
@@ -30,87 +29,96 @@
...
 
   @Autowired
-  private CamundaOperateClient camundaOperateClient;
+  public ProcessDefinitionImporter(
+      InboundConnectorRegistry registry, CamundaOperateClient camundaOperateClient) {
+    this.registry = registry;
+    this.camundaOperateClient = camundaOperateClient;
+  }
 
   @Scheduled(fixedDelay = 5000)
```
igpetrov, line 61: Not for now, but as an enhancement follow up. We should make this configurable.

igpetrov (approved):
LGTM. Thank you.

inbound/src/main/java/io/camunda/connector/inbound/importer/ProcessDefinitionImporter.java
igpetrov: Feel free to resolve.

berndruecker: I agree and added a TODO for now

igpetrov (approved):
Thank you for taking this.

PR #190, opened 2022-10-25
feat(hmac): integrated HMAC verification

igpetrov:
## Description

Integrates HMAC verification for webhook request.

## Testing done

Inbound connector: localhost
Camunda cluster: producton

Screenshot 2022-10-25 at 17 50 47

```
$> curl -XPOST -H 'Content-Type: application/json' -H "X-Hub-Signature-256: sha256=98ae3cdd258e3f7e7334d7963c779237392d05d32e991a167f3943e9f8747de2" localhost:8080/inbound/GITHUB_INBOUND  --data @example/webhook-payload-activates-packed.json
{"unauthorizedConnectors":[],"unactivatedConnectors":[],"executedConnectors":{"webhook-GITHUB_INBOUND-Process_PULL_REQUEST_NOTIFICATION-1":{"processDefinitionKey":2251799813685328,"bpmnProcessId":"Process_PULL_REQUEST_NOTIFICATION","version":1,"processInstanceKey":2251799813685336}},"errors":[]}

$> curl -XPOST -H 'Content-Type: application/json' -H "X-Hub-Signature-256: sha256=f93e172fb95efbc1406aee1140d6283f37a36b832cd4efc8b1e64114f29c86f3" localhost:8080/inbound/GITHUB_INBOUND  --data @example/webhook-payload-activates-packed.json
{"unauthorizedConnectors":["webhook-GITHUB_INBOUND-Process_PULL_REQUEST_NOTIFICATION-1"],"unactivatedConnectors":[],"executedConnectors":{},"errors":[]}
```

Screenshot 2022-10-25 at 17 52 14

## Related issues

- https://github.com/camunda/product-hub/issues/174
- https://github.com/camunda/connector-sdk-inbound-webhook/issues/11


igpetrov:
Hi @berndruecker,

Could you please take a look at this one at your convenience?

inbound/README.md
```diff
@@ -81,16 +95,13 @@ mvn spring-boot:run
 | Property | Required | Default value | Description |
 | :- | :- | :- | :- |
 | inbound.type | yes | | Needs to be set to `webhook` |
-| inbound.context | yes |  | Contxt path used on Webhook REST endpoint (http://endpoint/inbound/`context`/) |
+| inbound.context | yes |  | Context path used on Webhook REST endpoint (http://endpoint/inbound/`context`/) |
 | inbound.activationCondition | yes | |
-| inbound.variableMapping | no | |    
-| **Secret Handling** |
-| inbound.secretExtractor  | yes | |
```
berndruecker, line 88: Why is "secret" and "secretExtractor" removed? Do we not have this additionally?

inbound/example/.camunda/element-templates/generic-webhook-with-hmac.json
```diff
@@ -0,0 +1,136 @@
...
+      "group": "endpoint",
+      "optional": true,
+      "binding": {
+        "type": "zeebe:property",
+        "name": "inbound.hmacSecret"
+      }
+    },
+    {
+      "label": "HMAC Header",
+      "description": "Header that contains HMAC value",
```
berndruecker, line 75: Maybe better "Name of header attribute that will contain the HMAC value"? If we think people know what this is :-)

inbound/example/.camunda/element-templates/generic-webhook-with-hmac.json
```diff
@@ -0,0 +1,136 @@
...
+    },
+    {
+      "label": "Should validate HMAC",
+      "group": "endpoint",
+      "description": "Choose whether HMAC has to be validated",
+      "value": "enabled",
+      "type": "Dropdown",
+      "choices": [
+        {
+          "name": "Enabled (recommended)",
```
berndruecker, line 49: I am not sure if I would really make it a recommendation here in the modeler. E.g. in demos or prototypes I could imagine people are not doing it - but this then feels like it does not make too much sense.

inbound/src/main/java/io/camunda/connector/inbound/webhook/InboundWebhookRestController.java
```diff
@@ -127,7 +116,7 @@ private boolean isValidHmac(final WebhookConnectorProperties connectorProperties
                               final byte[] bodyAsByteArray,
                               final Map<String, String> headers)
           throws NoSuchAlgorithmException, InvalidKeyException {
-    if ("disabled".equals(connectorProperties.shouldValidateHMAC())) {
+    if (HMACSwitchCustomerChoice.disabled.name().equals(connectorProperties.shouldValidateHMAC())) {
```
berndruecker, line 119: Probably just a matter of taste - I like string constants in this case more than enums. But feel free to keep it like it is - will not make a difference any way

inbound/src/main/java/io/camunda/connector/inbound/webhook/InboundWebhookRestController.java
```diff
@@ -183,17 +172,4 @@ private boolean checkActivation(
     return Boolean.TRUE.equals(shouldActivate);
 //      throw fail("Failed to check activation", connectorProperties, exception);
   }
-
-  private boolean validateSecret(
```
berndruecker, line 187: Ok - so just to understand it correctly: HMAC is the way to provide secret management - there is no other way left?

berndruecker (approved):

berndruecker:
Sorry for the delay! I added some thoughts - but minor things or questions - I think the code is good to be merged.

inbound/README.md
igpetrov: I think, @nikku was assuming that we pass the secret passcode in header. However, this is not how it works. Instead of the passcode, the webhook caller passes a hash value.
Regarding the secret extractor - it was used to extract the secret passcode from the header. I moved this functionality into code and insted the customer has to specify only the header name where the hash value is stored.

inbound/src/main/java/io/camunda/connector/inbound/webhook/InboundWebhookRestController.java
igpetrov: Let me start with clarification, because secret is overloaded term and for the future reference for people in the future.

Secret here means a webhook token that concatenated with the content. It is defined when user creates a webhook. See GH example:

Screenshot 2022-10-26 at 18 09 36

The HMAC value is calculated as following:

`HMAC_VALUE = hash(algo, content, secret)`. That will produce a salted hash value, e.g. `sha256=aabbccdd`. This values is passed together with a header, e.g. `X-Hub-Signature-256`.

Initially, it was implemented under impression that secret will be passed in header but not the hash value. The secret token has to be put by user manually in the field `inbound.hmacSecret` so that we recalculate the content hash and validate with one that arrived.

igpetrov: So, yes, we don't plan to do anything in addition to validate webhook secret passcode, however, we need to add `secrets.*` handling logic so that user can store their secret in the Google Secrets Manager for SaaS or load them from the runtime / OS env - this logic is still pending. I am thinking, whether we can borrow it from the SDK.

inbound/example/.camunda/element-templates/generic-webhook-with-hmac.json
igpetrov: Makes sense. Updated.

inbound/example/.camunda/element-templates/generic-webhook-with-hmac.json
igpetrov: Updated.

inbound/src/main/java/io/camunda/connector/inbound/webhook/InboundWebhookRestController.java
berndruecker: Understood . make sense! Yes - we should definitely align on the Google Secrets Manager logic - so reuse the SecretsHandler/SecretsProvider from the connector-core. 
This way we can plug in different implementations (Google Secrets Manager for SaaS, Spring Environment for default runtime, probably custom Vault for self-managed, ...). 

As I plan to merge the ConnectorBridge into the Runtime soon - this should be the same code for Inbound/Outbound then. Do you want to move this to another issue and take the lead? Or should I? Should probably not block merging this one.

igpetrov: Sorry, didn't get what we want to split. I thought, we should merge this one first and develop secrets handling in a separate issue, which I can take. If you would like to merge ConnectorBridge in the middle or before this PR - please let me know.

berndruecker: Agree! Tag me in the issue - I can elaborate there (or ideally push some code to show what I mean :-))

berndruecker (approved):

berndruecker:
@igpetrov I think you can go ahead and merge your changes

inbound/src/main/java/io/camunda/connector/inbound/webhook/InboundWebhookRestController.java
berndruecker: Created Issue here: https://github.com/camunda/connector-sdk/issues/194

berndruecker:
Created issue to track the inbound secrets: https://github.com/camunda/connector-sdk/issues/194

PR #191, opened 2022-10-26
chore(naming): Renamed connector registration helper and extracted DTO

berndruecker:
I want to rename those classes as a prep step for the spring zeebe runtime (there it feels very unnatural to use the Registration itself as helper and value object at the same time).

I would love to merge this quickly and release some version of the connector-sdk to make progress with a change in spring-zeebe needed for the joined runtime (working locally on SNAPSHOTS already).

@tmetzke OK for you to merge? Could we then trigger a release? Either a 0.2.3 or some alpha if you prefer?


runtime-util/src/main/java/io/camunda/connector/runtime/util/outbound/OutboundConnectorValue.java
```diff
@@ -0,0 +1,107 @@
...
+  public OutboundConnectorFunction getFunction() {
+    return function;
+  }
+
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) return true;
+    if (o == null || getClass() != o.getClass()) return false;
+    OutboundConnectorValue that = (OutboundConnectorValue) o;
+    return Objects.equals(type, that.type);
```
tmetzke, line 84:
```suggestion
    if (this == o) {
      return true;
    }
    if (o == null || getClass() != o.getClass()) {
      return false;
    }
    OutboundConnectorValue that = (OutboundConnectorValue) o;
    return Objects.equals(type, that.type);
```

runtime-util/src/main/java/io/camunda/connector/runtime/util/outbound/OutboundConnectorValue.java
```diff
@@ -0,0 +1,107 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.camunda.connector.runtime.util.outbound;
+
+import io.camunda.connector.api.outbound.OutboundConnectorFunction;
+import java.util.Arrays;
+import java.util.Objects;
+
+public class OutboundConnectorValue {
```
tmetzke, line 23: I find that name a bit odd and generic, to be honest. Any thoughts on `OutboundConnectorConfiguration` maybe?
```suggestion
public class OutboundConnectorConfiguration {
```

tmetzke (changes requested):
Looks good to me as a change in general. Just one naming question and one code style issue.

Regarding a release: why exactly would we need one, now? The complication is, we cannot build alphas in an automated way currently, I have it on my list to adjust the release workflow to accommodate that. Building a patch like `0.2.3` would only make sense if we backport this PR to branch `stable/0.2`, which we could do in my regard if we really NEED a release.

**Edit**: Saw your explanation now. I can try to modify the release script now and see if I can allow alphas easily (there is some logic around building maintenance branches and stuff).

runtime-util/src/main/java/io/camunda/connector/runtime/util/outbound/OutboundConnectorValue.java
berndruecker: Don't have a strong feeling on this. "Value" is how those things are named in spring-zeebe - but I also like the one you propose. Happy to adjust it.

runtime-util/src/main/java/io/camunda/connector/runtime/util/outbound/OutboundConnectorConfiguration.java
```diff
@@ -0,0 +1,107 @@
...
+  public OutboundConnectorFunction getFunction() {
+    return function;
+  }
+
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) return true;
+    if (o == null || getClass() != o.getClass()) return false;
+    OutboundConnectorConfiguration that = (OutboundConnectorConfiguration) o;
+    return Objects.equals(type, that.type);
```
tmetzke, line 84:
```suggestion
    if (this == o) {
      return true;
    }
    if (o == null || getClass() != o.getClass()) {
      return false;
    }
    OutboundConnectorConfiguration that = (OutboundConnectorConfiguration) o;
    return Objects.equals(type, that.type);
```

runtime-util/src/main/java/io/camunda/connector/runtime/util/outbound/OutboundConnectorConfiguration.java
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(type);
+  }
+
+  @Override
+  public String toString() {
+    return "OutboundConnectorValue{"
```
tmetzke, line 98:
```suggestion
    return "OutboundConnectorConfiguration{"
```

tmetzke (changes requested):
Meh, sorry, I forgot that we have a class `OutboundConnectorConfiguration` already. This looks foolish, to be honest. Can we settle to stay with `OutboundConnectorRegistration`? I think that's actually fitting. Let's just move the utility methods into `OutboundConnectorRegistrationHelper`.

berndruecker:
@tmetzke Good catch! Question: Why don't we simply extent that class with the possibility to store the function and reuse it? Looking at the class now I think that make sense content-wise.

tmetzke (changes requested):
I think merging makes sense, but the helper/parser methods look a bit unnecessarily complex at first sight (creating a config out of a config). Would you mind looking at that?

If this takes too much time, I'd argue to revisit this later and live with the current "inconvenience" that it's one class that's a DTO and a utility.

berndruecker:
Haven't touched the parsing logic. Let's revisit that at a later point in time

berndruecker:
So if you are OK @tmetzke  I would merge it

runtime-util/src/main/java/io/camunda/connector/runtime/util/outbound/OutboundConnectorRegistrationHelper.java
```diff
@@ -144,14 +120,21 @@ private static OutboundConnectorRegistration parseConnector(final String name) {
...
+                        io.camunda.connector.impl.outbound.OutboundConnectorConfiguration::getType))
             .orElseThrow(() -> envMissing("Type not specified", name, "TYPE")),
         getEnv(name, "INPUT_VARIABLES")
             .map(variables -> variables.split(","))
-            .or(() -> config.map(OutboundConnectorConfiguration::getInputVariables))
+            .or(
+                () ->
+                    config.map(
+                        io.camunda.connector.impl.outbound.OutboundConnectorConfiguration
+                            ::getInputVariables))
```
tmetzke, line 137:
```suggestion
            .or(() -> config.map(OutboundConnectorConfiguration::getInputVariables))
```

tmetzke (approved):
Let's merge this then.

PR #208, opened 2022-10-27
chore: add more supported third-party licenses

### LINKED ISSUE #207: Add support for new third party licenses
We need to accommodate the following new dependencies in [our POM](https://github.com/camunda/connector-sdk/blob/main/parent/pom.xml#L374-L381) and/or [licenses template](https://github.com/camunda/connector-sdk/blob/main/licenses/third-party.ftl):
* Eclipse Public License - v 1.0 (shorten in POM and add to licenses template)
* GNU Lesser General Public License (shorten in POM and add to licenses template)
* Apache License (merge in POM)
* EPL 2.0 (merge in POM)

_Originally posted by @tmetzke in https://github.com/camunda/connector-sdk/issues/204#issuecomment-1293522310_

tmetzke:
## Description

Adds support for the following licenses in our TPL file:

* Apache License
* Eclipse Public License v. 1.0
* EPL-1.0
* Eclipse Public License - v 1.0
* EPL 2.0
* GNU Lesser General Public License, Version 2.1
* GNU Lesser General Public License

## Related issues

closes #207

### DIFF
--- a/licenses/third-party.ftl
+++ b/licenses/third-party.ftl
@@ -17,8 +17,12 @@
     <#return "https://opensource.org/licenses/BSD-3-Clause">
   <#elseif license == "EPL-2.0">
     <#return "https://www.eclipse.org/legal/epl-2.0/">
+  <#elseif license == "EPL-1.0">
+    <#return "https://www.eclipse.org/legal/epl-v10.html">
   <#elseif license == "GPLv2 with Classpath Exception">
     <#return "https://www.gnu.org/software/classpath/license.html">
+  <#elseif license == "GNU Lesser General Public License, Version 2.1">
+    <#return "http://www.gnu.org/licenses/lgpl-2.1.html">
   <#elseif license == "CDDLv1.0">
     <#return "https://opensource.org/licenses/CDDL-1.0">
   <#elseif license == "CDDLv1.1">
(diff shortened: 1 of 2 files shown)


igpetrov (approved):
LGTM. One observation from my experience is that corporate users might not want to use these licenses, especially GNU because code they extend has to become open-source which might not be acceptable in case, users introduce their mission critical code.

tmetzke:
I am currently adding a guide on dependencies. All of the libraries we use must come with a license that does not enforce what you just outlined 👍

berndruecker:
We might simply make it clear to people that this is not an issue with a logging library that can be swapped out. 
It might be a matter of how we communicate this. E.g. Spring Boot is also clearly Apache License (and come with this logging dependency as default - but anybody could simply exchange it).

tmetzke:
Well, yes and no, I think. We are not only providing a Spring Boot Starter project here that people use as dependency and can adjust. We are distributing a pre-packaged application that contains the library. Users cannot simply remove the library from that. That is also why we need to pay attention at what we are consuming and distributing.

berndruecker:
Legal gave a thumbs up - so it is not a problem. 
"Are we linking dynamically (as opposed to statically): If yes, good.
Are we NOT modifying (e.g., making a derivative) the code? If yes, good."

If we somehow experience any acceptance problems with EPL we could still exclude backlog and hook in slf4j instead.

tmetzke:
Thanks for following up, @berndruecker 👍
Both aspects I would also answer with a YES 👍

PR #211, opened 2022-10-27
docs: fix dead link

vil02:
## Description

This PR fixes the dead link in [README.md](https://github.com/camunda/connector-sdk/blob/93e3591ceb0981e123bca76164bad49181ddf306/README.md?plain=1#L84). I think the reason for it was the recent renaming of the directory `runtime-job-worker` to `runtime` (cf. PR #188 ).

## Related issues

No related issues.

PR #212, opened 2022-10-28
docs: fix broken link to `runtime`

vil02:
## Description

This PR fixes the broken link in [`./core/README.md`](https://github.com/camunda/connector-sdk/blob/c3e7b0828458774660a97ec6033295d74547841e/core/README.md?plain=1#L34). The reason of the problem is the same as described in PR #211 (I did not spot it then).

## Related issues

There no related issue.


tmetzke (approved):
Thanks, @vil02, I just saw this as well and was about to fix this, but you're too fast 🐎 👍

vil02:
@tmetzke thank you for a quick merge!

PR #213, opened 2022-10-28
Outbound connector function dead link patch

vil02:
## Description

I updated the [`./core/README.md`](https://github.com/camunda/connector-sdk/blob/824754efb8efc48167f173504999355005690bd6/core/README.md), i.e.
- fixed the dead link in [`./core/README.md`](https://github.com/camunda/connector-sdk/blob/824754efb8efc48167f173504999355005690bd6/core/README.md?plain=1#L7),
- updated the example [`PingConnector`](https://github.com/camunda/connector-sdk/blob/824754efb8efc48167f173504999355005690bd6/core/README.md?plain=1#L9).

## Related issues

There is no related issue.


tmetzke (approved):
Awesome, thanks @vil02 for validating our docs here 👍 
Looks like we only adjusted this on the root README.

Issue #205, opened 2022-10-27
Document classloading with multiple connectors in one runtime


berndruecker:
See https://github.com/camunda/team-connectors/issues/125

Classloading problems already happened with `jackson`, but the more connectors we build, the more classloading issues we will see.

**Ideas**

- **Relocate** library classes via mvn shade plugin in connector fat jars
  - Could be configured by an obligatory parent pom
  - Maybe simply relocate all classes except connector API to make sure, we don't have overlaps
  - Con:
    - Every connector brings a copy of common libraries
    - package names (e.g. in exceptions) are "unnormal"
    - Connector developers need to stick to the rules
- **Align** the library versions for all connectors (at least for common libraries)
  - For example via parent pom or BOM
  - Con: 
    - Hard to update versions (would need a rebuild of all connectors to use the latest version)
    - Hard to control that connector developers really stick to it
- Don't build a jar with dependencies per connector but a single jar with dependencies bundled with all connectors (**Maven** does dependency conflict resolution)
  - Easy to do for our own OOTB bundle of connectors
  - Con:
     - Does not allow for a "DevOps"-way of using connectors (e.g. just stitch together a couple of jars in a Docker file
- Some Docker magic to do a mvn build on-the-fly
  -  We could provide a Docker file that can create a big Maven build with all provided connectors to produce its own uber JAR on-the-fly (I think)
  - Con:
     - Tied to docker
     - Could get us into some nasty trouble ;-)
- Build some **custom classloading** isolation 
  - Con:
    - Super complex
    - App servers and OSGI did always struggle with that
- Add some **remote invocation** to the connector code (like via Connector Bridge at the moment, or even RMI) to put connector into own JVM
  - Isolates connectors from the runtime much better, probably helpful also in other regards (thinking about running community connectors in a sandbox)
  - Con:
    - Added complexity in deployment
    - Less efficiency and more possible failure cases to handle

berndruecker:
Thoughts on **Fat Jar** vs **Maven Build** from @tmetzke:

Yes, building this with Maven is one way of working around this and we have considered this. We think this doesn't solve the issue for more DevOps-ish people that want to run/administer Connectors. They'd rather want to decide which Connectors to add to the runtime without handling Maven dependencies, but rather adding them to Docker images or k8s clusters. That's why we currently have the FAT JARs approach. So administrators can mix and match a pre-packaged runtime (that we provide as a Docker image) with any Connectors they'd like to run.

[...]

PR #215, opened 2022-10-28
Adjust `java -cp` command for Spring-based runtime

berndruecker:
We need to use the Spring JarLoader - but still looking into classloading overall (https://github.com/camunda/connector-sdk/issues/205).

We also need to adjust the `Dockerfile` and docs - so I wanted to wait if this is the final call.


tmetzke:
When I use the `JarLauncher` to start the runtime, there is no logging. It looks like it's not picked up:
```
SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.

  .   ____          _            __ _ _
 /\\ / ___'_ __ _ _(_)_ __  __ _ \ \ \ \
( ( )\___ | '_ | '_| | '_ \/ _` | \ \ \ \
 \\/  ___)| |_)| | | | | || (_| |  ) ) ) )
  '  |____| .__|_| |_|_| |_\__, | / / / /
 =========|_|==============|___/=/_/_/_/
 :: Spring Boot ::                (v2.7.5)
```

Any ideas on why that happens? Works fine with the `PropertiesLauncher` and `-Dloader.path`.

tmetzke:
We settled for #223 for now, using `maven-shade-plugin` instead of `spring-boot-maven-plugin`. 

The biggest problem is that using the `JarLauncher`, the dependencies that come with the über JAR Connectors are prioritized over those that come with the Connector Runtime. This leads to, for example, logging not working correctly.

Using the `PropertiesLauncher` and `-Dloader.path` that actually works (because the Spring Boot app is the driver and the first thing that loads, adding the Connectors to the classpath correctly for our use case).

We could also go back to the `PropertiesLauncher` approach but would have to rewrite the instructions and Dockerfile. Plus, we would have to say "please don't use this with `JarLauncher` because that doesn't work properly". I'd like to avoid that.

PR #223, opened 2022-11-02
chore(runtime): remove inbound until ready

### LINKED ISSUE #222: Get `runtime` ready for `0.3.0`
__What should we do?__

- Exclude Inbound from the `runtime` for `0.3.0` since Inbound is not ready yet
- Make sure the Dockerfile and READMEs are aligned regarding starting the runtime (currently, they differ)

__Why should we do it?__

Prepare `runtime` so it's stable for `0.3.0`

tmetzke:
## Description

Prepares `runtime` for `0.3.0` release

## Related issues

closes #222


igpetrov (dismissed):

tmetzke:
Let's also fix the Dockerfile and README in here to make the `runtime` usable again 🙂

tmetzke:
Built the Docker container locally, I added the REST Connector manually from a local file.
Spring Boot application started, Connector was loaded and executed.

igpetrov (approved):

berndruecker:
OK - thanks @tmetzke - so far I haven't seen the variant of adding the Spring META-INF magic manually. I like that we end up with a normal JAR without classpath magic - even if I am a bit afraid of something breaking with Spring Updates - but let's wait for that and keep the current logic 👍

tmetzke:
We've done something similar in C7 a while back, that's why I knew you can simply pull this together "manually" if you like. In the end, the Spring Boot Maven plugin only adds its own loader logic, but it's not entirely necessary.

Issue #228, opened 2022-11-04
Get `inbound` ready for `0.3.0`


tmetzke:
__What should we do?__

Overall, we need to ensure `inbound` (although still a work in progress) is coherent and can be used (we can also list limitations, if necessary). Plus, `inbound` should not introduce any regressions/unexpected behavior for `outbound` users by default (e.g. creating a webhook endpoint, spin up Operate importers and poll).

- [x] Validate `inbound` README (currently, the described procedures don't work from inside the module or are unclear where to be run from)
- [x] Ensure including `inbound` in `runtime` does not spin up resources by default (webhook endpoint, Operate polling, ...) --> https://github.com/camunda/connector-sdk/issues/232
- [x] Remove Spring Zeebe from `inbound` (should not be necessary), it's provided by `runtime`.

__Why should we do it?__

Get `inbound` release-ready for `0.3.0`.

tmetzke:
## Reasoning

As far as I can see, we want to have the following in the long run:
- Runtime contains Spring Zeebe
- Runtime has a main Spring Boot class
- Runtime includes outbound (via Spring Zeebe) and inbound (via inbound) capabilities
- You can configure what to start externally (outbound connector functions, inbound connector webhook endpoint, ...)

But I think we're not there yet. Too much stuff simply starts by default. It might make sense to give ourselves more time to get this right. Thus, we could currently keep `runtime` as the Outbound runtime and keep `inbound` self-contained and separate from `runtime` until we're ready (e.g. `0.4.0`).

berndruecker:
Working on it in PR https://github.com/camunda/connector-sdk/pull/231 (draft)

berndruecker:
Would you agree this can be closed @tmetzke? 
Also related to https://github.com/camunda/connector-sdk/issues/202 which is "close to be closed" (waiting for some PR's to be merged) :-)

PR #231, opened 2022-11-07
Prepping runtime for 0.3.0

berndruecker:
PR for https://github.com/camunda/connector-sdk/issues/228


igpetrov (dismissed):
I am curious, why do we need separate flags for webhook and polling?

It feels like, when webhook is on and polling off, it won't work anyway.

berndruecker:
The other way round - I would find it logical to enable polling for inbound connectors, but switch webhooks off (because I only want to have inbound JMS for example). This way the properties should keep stable

igpetrov:
> The other way round - I would find it logical to enable polling for inbound connectors, but switch webhooks off (because I only want to have inbound JMS for example). This way the properties should keep stable

Understood, thank you

berndruecker:
@tmetzke Ready to be merged. 
This reintroduces `inbound` to the runtime, but disable side effected by the default properties (polling, webhook, Tomcat). 
Adjusted readmes.

inbound/README.md
```diff
@@ -121,7 +62,7 @@ zbctl --insecure deploy resource example/pull-request-notification-hmac-on.bpmn
 zbctl --insecure deploy resource example/broken.bpmn
 ```
 
-### Start a process
+###' Start a process
```
igpetrov, line 65:
```suggestion
### Start a process
```

igpetrov (approved):
LGTM

igpetrov (approved):

tmetzke (commented):
Nice, thanks @berndruecker 👍 
Happy to merge the runtimes again like this 🙂

tmetzke (approved):

PR #233, opened 2022-11-07
chore(runtime): include validation module

tmetzke:
## Description

Includes the validation module by default in the `runtime`. This provides the module so Connectors can consume it if they want to. Connectors (optionally) depend on the validation module with scope `provided`. Thus, as their runtime, we need to provide it here.

## Related issues

related to https://github.com/camunda/team-connectors/issues/209

### DIFF
--- a/runtime/pom.xml
+++ b/runtime/pom.xml
@@ -19,6 +19,10 @@
       <artifactId>connector-runtime-util</artifactId>
     </dependency>
     <dependency>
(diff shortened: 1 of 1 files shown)


igpetrov (approved):
Perfect. Thank you. Now, `provided` should work, isn't it?

tmetzke:
> Perfect. Thank you. Now, `provided` should work, isn't it?

Yes, exactly. 👍

PR #238, opened 2022-11-07
feat(core): allow secret in any position

### LINKED ISSUE #236: Allow secrets in any position of a string
__Define the problem__

As a user, I want to be able to concatenate a secret with another string to be able to access REST APIs that offer a custom authentication format (e.g. the GitHub API that expects `token <TOKEN>` as a value of the `Authorization` header, see for example https://docs.github.com/en/rest/issues/issues#create-an-issue).

This doesn't work with the current implementation because secrets are only replaced if a string only contains a secret and nothing else:

https://github.com/camunda/cloud-connector-http-json/blob/9950be85911e0ad70bc4fd29d51df54561956a87/src/main/java/io/camunda/connector/http/SecretStore.java#L21

__Describe the solution__

* Allow secrets in any position of a string and not just replace it if the string only contains the secret and nothing else. 
* Document the current limitation in the meantime as users might not be aware of it

tmetzke:
## Description

Allows to use secrets anywhere in a String. Either the whole String represents a secret as `secrets.<SECRET_NAME>` or secrets are wrapped as placeholders like `Text with secret: {{secrets.SECRET_NAME}}`. For the latter, the whole placeholder `{{secrets.SECRET_NAME}}` will be replaced by the secret value.

## Related issues

closes #236


core/src/test/java/io/camunda/connector/impl/outbound/OutboundConnectorContextTest.java
```diff
@@ -449,6 +451,43 @@ void shouldReplaceSecretsStringSetWithCustomHandler() {
           .allMatch(
               s -> "plain".equals(s) || "plain".equals(((OutboundTestInput) s).getSecretField()));
     }
+
+    @ParameterizedTest
+    @ValueSource(strings = {"secrets.FOO", "{{secrets.FOO}}", " secrets.FOO ", " {{secrets.FOO}} "})
```
nikku, line 456:
```suggestion
    @ValueSource(strings = {"secrets.FOO", "{{secrets.FOO}}", " secrets.FOO ", " {{secrets.FOO}} ", "{{ secrets.FOO}}", "{{secrets.FOO }}"})
```

tmetzke: implemented with [b637e61](https://github.com/camunda/connector-sdk/pull/238/commits/b637e61c2c6a85c84d1b420962863e87689f276c)

core/src/main/java/io/camunda/connector/api/secret/SecretStore.java
```diff
@@ -27,7 +28,9 @@
  */
 public class SecretStore {
 
-  private static final Pattern SECRET_PATTERN = Pattern.compile("^secrets\\.(\\S+)$");
+  private static final Pattern SECRET_PATTERN_FULL = Pattern.compile("^secrets\\.(?<secret>\\S+)$");
+  private static final Pattern SECRET_PATTERN_PLACEHOLDER =
```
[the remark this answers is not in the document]
tmetzke: implemented with [b637e61](https://github.com/camunda/connector-sdk/pull/238/commits/b637e61c2c6a85c84d1b420962863e87689f276c)

tmetzke:
> > Consider making the pattern a little bit more forgiving (add `\s*` before and after the secrets string).
> 
> Very nice catch!! Thanks, will implement! 👍

Done with [b637e61](https://github.com/camunda/connector-sdk/pull/238/commits/b637e61c2c6a85c84d1b420962863e87689f276c)

PR #241, opened 2022-11-08
Removing SecretStore as it does not do anything meaningful

berndruecker:
The SecretStore might be remaining code from earlier iterations (e.g. lazy loading of secrets). At the moment it does not do anything and is actually confusing (what is the SecretProvider vs. the SecretStore).

When looking at SecretProviders I found that we should remove it. Better now than later. It should not affect connector implementions as they should simply use `replaceSecrets` in the ConnectorContext.

OK @igpetrov @tmetzke ? 

PS: Would need changes in spring-zeebe as well. Now I stumble over the cyclic release dependency - but happy to resolve that if I have some green light :-)


igpetrov:
Good call. I was confused the other day between `SecretStore`, `SecretProvider`, and `SecretHandler`.

Seems you may have a merge conflict with https://github.com/camunda/connector-sdk/pull/238/files

berndruecker:
Happy to merge changes when the other PR is on main - no problem (was not a big deal either :-))

tmetzke:
The `SecretStore` contains the main routine to replace secrets. Users should not touch this ideally. The `SecretProvider` is the SPI that people can implement to connect to their external secret manager, to pull in the secret value for a specific name/key. The `SecretHandler` determines how to handle container objects that can contain secrets. All of them should actually stay, in my point of view.

tmetzke:
I'll have a closer look now, could only look at this yesterday in between meetings.
My gut feeling still is that we need those classes separated. We can definitely improve internal documentation and/or rename classes if that helps.

tmetzke:
I think we can get rid of the `SecretStore` class indeed. I am working out some minor details right now and will probably file a new PR, superseding this one here. Thanks for bringing this up, @berndruecker 👍 
I'll probably close this one here once I have the superseding one, if that's fine with you, @berndruecker?

berndruecker:
yeah . fine with me

tmetzke:
superseded by #246

Issue #195, opened 2022-10-27
Resolve release dependency chain `runtime` -> `spring-zeebe` -> `runtime-util`


berndruecker:
Currently we have a nasty cycle dependency, as spring-zeebe needs to depend on runtime-util, but the actual runtime needs spring-zeebe. This is a problem when releasing new features (as you might need to do a connector-sdk release, to enabne the spring-zeebe release which is needed to move forward in cnnectors.

However, I don't think we should change that now - as it is also pretty convenient to have everything in one build, especially for extending the SDK towards Inbound (e.g. Subscription & Polling).

But at some point in time I think we should split the current connector-sdk into:

- `connector-sdk`, containing the real SDK (`core`) and Utilities ('runtime-util') (no Spring dependency here)
- `connector-runtime` containing the runtime (spring-zeebe based stuff)

tmetzke:
Just to understand this:

We don't technically have a cyclic dependency issue here, right? That would mean, as far as I know, that `runtime` depends on `spring-zeebe` and `spring-zeebe` depends on `runtime`. 

This is rather a release issue, right? I agree with the proposal that we might need to extract the runtime from the SDK.

Then again, I was just wondering: 

* Can't Spring Zeebe simply BE the runtime?
* Do we actually need another `runtime` project at all?
* What benefit does it provide?
* Can we integrate the stuff that is needed into Spring Zeebe itself?

berndruecker:
I like the idea of Spring Boot being the runtime. 

We could basically move the current `runtime` to `spring-zeebe-connector-runtime` as submodule in spring Zeebe. The code from `inbound` might be distributed to either runtime-util or this new module. Some generic stuff might also end up in spring-zeebe itself - not yet sure. 

Connector-SDK would "only" be:

- connector-core
- connector-validation
- runtime-util 

This way, connector developer would look towards https://github.com/camunda/connector-sdk and runtime folks towards https://github.com/camunda-community-hub/spring-zeebe + `/connector-runtime`?

Would https://github.com/camunda/connectors-bundle stay an independant repo?

tmetzke:
That sounds a lot like my vision for this, yes 👍 
The SDK itself would just bring along the basic components to build Connectors (exactly as you detailed it).
Spring Zeebe would be the default runtime that we deliver.

Connectors Bundle delivers a whole "distribution" including element templates and backend (runtime + Camunda-provided Connectors). If we want something like this inside Spring Zeebe, let's add it. 
[...]

PR #246, opened 2022-11-09
chore!: merge `SecretStore` into `SecretHandler`

### LINKED ISSUE #245: Clean up secrets handling classes
__What should we do?__

Merge the `SecretStore` into the `SecretHandler` and remove it from API.

__Why should we do it?__

Clean up the classes responsible for secret handling.

tmetzke:
## Description

* Fixes the secret replacement routine in the `SecretHandler` so it can be reused by all Connector context implementations easily. The SecretHandler becomes the recommended class to use for replacing secrets.

BREAKING CHANGE: The class `SecretStore` is removed, use the `SecretHandler`.

## Related issues

closes #245


core/src/main/java/io/camunda/connector/impl/secret/SecretHandler.java
```diff
@@ -41,10 +44,14 @@ public class SecretHandler implements SecretElementHandler, SecretContainerHandl
   protected static final List<Class<?>> PRIMITIVE_TYPES =
       List.of(String.class, Number.class, Boolean.class);
 
-  private final SecretStore secretStore;
+  private static final Pattern SECRET_PATTERN_FULL = Pattern.compile("^secrets\\.(?<secret>\\S+)$");
+  private static final Pattern SECRET_PATTERN_PLACEHOLDER =
+      Pattern.compile("\\{\\{\\s*secrets\\.(?<secret>\\S+?\\s*)}}");
```
igpetrov, line 49: Sorry, not a regex expert but I find weird that first `{` are escaped but the last `}` are not. Is it expected?

tmetzke:
@berndruecker, I had to disable some tests due to them using Spring Zeebe for testing. The latest version however still uses the `SecretStore`, which is why the tests cannot startup. Should be solved with #195.

core/src/main/java/io/camunda/connector/impl/secret/SecretHandler.java
tmetzke: Yeah, this is actually just a Java regex issue. In a regex, you actually don't need to escape them (but you can). In Java, if a regex String contains `{`, this is considered as starting a quantification like `{3}`. If you want `{` to be taken literally, you have to escape it in Java regexes. For `}`, the Java regexes are clever enough to see there is no `{` starting a quantification, so you don't have to escape them.

igpetrov (approved):
LGTM

Issue #243, opened 2022-11-08
Harmonize secret handling and make Google Secret Manager pluggable


berndruecker:
Follow up on https://github.com/camunda/connector-sdk/issues/194

SDK: https://github.com/camunda/connector-sdk/tree/main/core/src/main/java/io/camunda/connector/api/secret

And e.g. in Spring I have an implementation simply going to the Spring Environment: https://github.com/camunda-community-hub/spring-zeebe/blob/master/client/spring-zeebe/src/main/java/io/camunda/zeebe/spring/client/jobhandling/SpringSecretProvider.java

And here is the current implementation going against the Google Security Manager: https://github.com/camunda/connector-runtime-cloud/blob/main/src/main/java/io/camunda/connector/runtime/cloud/CloudSecretProvider.java, this needs to go somewhere, where we can easily use it by cofniguration

- [x] Move Google Security Manager Code to SDK --> https://github.com/camunda/connector-sdk/pull/251
- [x] Chose Security Provider via Spring config (& env variables, ...) if multiple are found

having both will allow to use the normal Spring-based runtime together with the Google Security Manager by just setting the right flag in the config.

berndruecker:
Moved the basic code for GCP here: https://github.com/camunda/connector-sdk/pull/251

berndruecker:
A cache is used - it reloads data if older than 5 seconds.

Still need to be tested E2E

PR #251, opened 2022-11-11
feat(security): Provide GCP Security Manager as Secret Provider

berndruecker:
Related to https://github.com/camunda/connector-sdk/issues/243

I copied the code from https://github.com/camunda/connector-runtime-cloud/blob/main/src/main/java/io/camunda/connector/runtime/cloud/CloudSecretProvider.java as a separate module into the SDK (I think it is better suited here than close to the runtime)

Not yet used - but I when I start removing the cloud-bridge we will need this in the runtime soon


nikku:
Shall this be public?

secret-provider/gcp-security-manager/src/main/java/io/camunda/connector/runtime/cloud/CloudSecretProvider.java
```diff
@@ -0,0 +1,77 @@
...
+
+  private static String loadGoogleSecrets(final String clusterId) {
+    LOGGER.info("Fetching secrets for cluster {} from secret manager", clusterId);
+    try (final SecretManagerServiceClient client = SecretManagerServiceClient.create()) {
+      final String projectId =
+          Objects.requireNonNull(
+              System.getenv(SECRETS_PROJECT_ENV_NAME),
+              "Environment variable " + SECRETS_PROJECT_ENV_NAME + " is missing");
+      final String secretPrefix =
+          Objects.requireNonNullElse(System.getenv(SECRETS_PREFIX_ENV_NAME), "connector-secrets");
```
tmetzke, line 48: I'd go with something similar to the line above. This must be provided or fail, we shouldn't have a default in the code here.

secret-provider/gcp-security-manager/src/main/java/io/camunda/connector/runtime/cloud/CloudSecretProvider.java
```diff
@@ -0,0 +1,77 @@
...
+
+  private final Map<String, String> secrets;
+
+  public CloudSecretProvider(final Gson gson, final String clusterId) {
+    final String json =
+        Optional.ofNullable(clusterId)
+            .map(CloudSecretProvider::loadGoogleSecrets)
+            .or(CloudSecretProvider::loadEnvironmentSecrets)
+            .or(CloudSecretProvider::loadPropertiesSecrets)
+            .orElse("{}");
```
tmetzke, line 33: I think we need to fix this. Either the secrets are loaded from the Secret Manager or this fails.

secret-provider/gcp-security-manager/src/main/java/io/camunda/connector/runtime/cloud/CloudSecretProvider.java
```diff
@@ -0,0 +1,77 @@
...
+      final String secretName = String.format("%s-%s", secretPrefix, clusterId);
+      final SecretVersionName secretVersionName =
+          SecretVersionName.of(projectId, secretName, "latest");
+      final AccessSecretVersionResponse response = client.accessSecretVersion(secretVersionName);
+      return response.getPayload().getData().toStringUtf8();
+    } catch (final Exception e) {
+      LOGGER.info(
+          "Failed to load secrets from secret manager, falling back to environment secret store",
+          e);
+      return null;
```
tmetzke, line 59: I think we need to fix this. Either the secrets are loaded from the Secret Manager or this fails.

secret-provider/gcp-security-manager/pom.xml
```diff
@@ -0,0 +1,94 @@
...
+        </plugin>
+        <plugin>
+          <groupId>com.mycila</groupId>
+          <artifactId>license-maven-plugin</artifactId>
+          <configuration>
+            <skip>true</skip>
+          </configuration>
+        </plugin>
+      </plugins>
+    </pluginManagement>
```
tmetzke, line 92:
```suggestion
```

secret-provider/gcp-security-manager/pom.xml
```diff
@@ -0,0 +1,94 @@
...
+  </build>
```
tmetzke, line 93:
```suggestion
```

tmetzke:
> Shall this be public?

I think the code in general can be public, there is no secret/internal information in here, is there?

nikku:
Just wanted to make sure. Feels odd to expose our internals (public or not) as part of the SDK.

berndruecker:
I agree with @tmetzke - no sensitive information given here. And if there is an issue how we use GCP I prefer the community to spot it openly so we fix it faster (but not concerned here at all honestly).

i was not improving the code at all at this stage - simply moving it.

tmetzke:
> I was not improving the code at all at this stage - simply moving it.

Let's try to get the easy fixes out of the door directly here. I can also tackle that next week, if necessary.

secret-provider/gcp-security-manager/src/main/java/io/camunda/connector/runtime/cloud/CloudSecretProvider.java
berndruecker: This is a good hint. Actually it does not make sense overall for the CloudSecretProvider to load secrets from anywhere but Google Cloud Secret Manager. I will remove the other  options and let it fail it cannot get things from Google. 

When I want to test something without Google but Environment Variables, I should use a different SecretProvider.

secret-provider/gcp-security-manager/src/main/java/io/camunda/connector/runtime/cloud/CloudSecretProvider.java
```diff
@@ -0,0 +1,73 @@
...
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(CloudSecretProvider.class);
+  private static final Type MAP_TYPE = new TypeToken<Map<String, String>>() {}.getType();
+  public static final String SECRETS_PROJECT_ENV_NAME = "SECRETS_PROJECT_ID";
+  public static final String SECRETS_PREFIX_ENV_NAME = "SECRETS_PREFIX";
+
+  private final Map<String, String> secrets;
+
+  public CloudSecretProvider(final Gson gson, final String clusterId) {
+    String json = loadGoogleSecrets(clusterId);
```
tmetzke, line 41: I am just wondering how we are using this provider. Usually, the provider is instantiated using the SPI mechanism. This requires a no-args constructor, as far as I know. So we cannot instantiate this one here.

Plus, even if this were possible here, the secrets are loaded once when the provider is initialized. This makes sense for our GCP functions because the context is created there on each function invocation. With Spring Zeebe, we however create the provider once, if I am not mistaken? That would mean that new secrets that were added after the initialization wouldn't be fetched. Or am I missing something here?

berndruecker: Let's see how we initialize, currently looking into this (also proposed https://github.com/camunda/connector-sdk/pull/256 a minute ago). I might adjust this here then (WIP).

You are right for eager-fetching not updating the secrets. We could change this to either
- be reloaded every x seconds
- every time this is used

PR #256, opened 2022-11-14
Prepare JobHandler to get SecurityProvider injected to remove logic here

berndruecker:
@tmetzke Would you be open to include this change before the 0.3.0 release?
It doesn't change anything but would allow to evolve spring-zeebe without requiring to base on a 0.4 alpha.


tmetzke:
Happy to add this in general. However, you can also achieve this by subclassing the handler and overriding the `getSecretProvider` like I did in Spring Zeebe, can't you? What is the benefit here? Not having to subclass the handler?

If we want to have this, I'd like to make this a bit more sound. Can we rename the `getSecretProvider()` method to `loadSecretProviderSPI()`? I think that describes better what it does then.

tmetzke:
Ah no, damn it, this will mean all users of this class will have to be adjusted as well. Meh.

tmetzke:
But that probably only is Spring Zeebe currently, I guess.

berndruecker:
I think it is bad design if the Job Handler creates a Security Provider itself under the hood, it should always get it from the environment. Typically, also inbound and outbound will share the SecurityProvider. 

I would do the current change to prepare the removal of the getSecurityProvider() completely, this will only affect spring-zeebe then.

runtime-util/src/main/java/io/camunda/connector/runtime/util/outbound/ConnectorJobHandler.java
```diff
@@ -46,6 +47,12 @@ public class ConnectorJobHandler implements JobHandler {
    */
   public ConnectorJobHandler(final OutboundConnectorFunction call) {
     this.call = call;
+    this.secretProvider = getSecretProvider();
```
berndruecker, line 50: If we are afraid doing this not lazy, we might do lazy as intermediary.

runtime-util/src/main/java/io/camunda/connector/runtime/util/outbound/ConnectorJobHandler.java
```diff
@@ -46,6 +47,12 @@ public class ConnectorJobHandler implements JobHandler {
...
+  }
+
+  public ConnectorJobHandler(final OutboundConnectorFunction call, SecretProvider secretProvider) {
```
igpetrov, line 53:
```suggestion
  public ConnectorJobHandler(final OutboundConnectorFunction call, final SecretProvider secretProvider) {
```

igpetrov (dismissed):

tmetzke:
I agree that it's nice to support providing the secret provider from the outside in general. 👍 

However, with this change, we definitely adjust behavior: we only initialize the provider once instead of on every function invocation. It is not as simple a change as it may seem at first sight.

berndruecker:
I don't see any situation where this could hurt as at the moment - but especially given that I want to remove the code soon anyway - I changed that behavior to shortcut any possible discussion right now :-) WDYT @tmetzke ?

tmetzke:
Suggestion at this point (because I am stuck with other release things and I think we can also fix this later): Let's do this properly after the release, regarding:
- Where should the provider come from?
- When should we initialize it?
- Who should decide how often it fetches freshly from the store of secrets?

We can crank out an alpha of `0.4.0` on Wednesday if this is solved then and we need it. We can release fast and often in the SDK, that should be no blocker, I think.

berndruecker:
Replaced by https://github.com/camunda/connector-sdk/pull/276

PR #262, opened 2022-11-21
Added ConnectorConfigurationHelper 

berndruecker:
First step to fix GCP Security Manager Code

Still WIP


secret-provider/gcp-security-manager/src/main/java/io/camunda/connector/runtime/cloud/CloudSecretProvider.java
```diff
@@ -31,15 +34,28 @@
 public class CloudSecretProvider implements SecretProvider {
 
   private static final Logger LOGGER = LoggerFactory.getLogger(CloudSecretProvider.class);
+  private static final Gson GSON = new GsonBuilder().create();
```
igpetrov, line 37: Suggestion: maybe we could initialize gson from constructor instead of static context. The user who initializes it will be responsible for their own configured gson instance.

In spring zeebe it will be taken from application context.

igpetrov (approved):
LGTM. Minor optional suggestion.

berndruecker:
Closing in favor of https://github.com/camunda/connector-sdk/pull/266

PR #278, opened 2022-11-24
feat: Allow to read outboundConnectorFunction in ConnectorJobHandler

berndruecker:
@tmetzke Need this to query some information for metrics in spring-zeebe


berndruecker:
I take that back ;-)

PR #279, opened 2022-11-24
Made GCP credential cache configurable

berndruecker:


core/src/main/java/io/camunda/connector/impl/config/ConnectorConfigurationUtil.java
```diff
@@ -41,4 +41,14 @@ public static String getProperty(String key) {
     }
     return System.getenv(key);
   }
+
+  public static String getProperty(String key, String defaultValue) {
+    String result = getProperty(key);
+    if (result==null) {
```
tmetzke, line 47:
```suggestion
    if (result == null) {
```

tmetzke (approved):
👍

PR #296, opened 2022-12-06
Allow Secret prefix to be configured

berndruecker:
required to reuse class for another prefix to load internal M2M Secrets (see https://github.com/camunda/team-connectors/issues/247)


tmetzke:
To make sure I get the context right:
- We want to have two instances of the Provider in one application (Connector Runtime).
- For the Connector secrets we inject the GCP Secret Provider as SPI and configure it via ENV vars.
- For the Operate connection we create the provider in code and hardwire the configuration values (or fetch them from different ENV vars than the defaults)?

tmetzke (approved):
Changes look good to me 👍

PR #334, opened 2023-01-10
added necessary classes for inbound/subscription

berndruecker:
Quick first stab at a possible inbound/subscription solution for discussion


berndruecker:
Will be done via https://github.com/camunda/connector-sdk/pull/346

PR #346, opened 2023-01-17
Inbound Intermediate Events (SDK changes)

chillleader:
## Description

SDK changes required for inbound intermediate events.
To be reviewed in bundle with...

## Related issues

https://github.com/camunda/team-connectors/issues/308


core/src/main/java/io/camunda/connector/api/inbound/InboundConnectorTarget.java
```diff
@@ -0,0 +1,46 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.camunda.connector.api.inbound;
+
+import java.util.Map;
+
+public abstract class InboundConnectorTarget implements Comparable<InboundConnectorTarget> {
```
berndruecker, line 21: Not a concern - but a quick feedback: I really wondered what a InboundConnectorTarget is until I looked into the class. Maybe we find a better name? Just quick ideas:

- ProcessEndpoint
- ProcessCorelationInfo
- ProcessInfo
- ProcessInvocationParameters

But as always - naming is hard - not a clear proposal :-/

core/src/main/java/io/camunda/connector/api/inbound/InboundConnectorWrapperException.java
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.camunda.connector.api.inbound;
+
+public class InboundConnectorWrapperException extends RuntimeException {
```
berndruecker, line 19: Again just a thought: Do we need an own Inbound exception or is a ConnectorException probably sufficient? As SDK user I think it is easier to have less exception classes - and in this case the differentiation has no big advantage

core/src/main/java/io/camunda/connector/api/inbound/InboundConnectorResult.java
```diff
@@ -0,0 +1,23 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.camunda.connector.api.inbound;
+
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
+import com.fasterxml.jackson.annotation.JsonTypeInfo.Id;
+
+@JsonTypeInfo(use = Id.NAME, property = "type")
```
berndruecker, line 22: THis is the only place in the SDK where we rely on Jackson / add the depenency to the JDK. Not against it - just wondering (maybe I understand the need when looking into Spring Zeebe :-))

berndruecker (commented):
I generally like the direction very much! I made some minor comments

core/src/main/java/io/camunda/connector/api/inbound/InboundConnectorProperties.java
```diff
@@ -0,0 +1,61 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.camunda.connector.api.inbound;
+
+import java.util.Map;
+
+public class InboundConnectorProperties {
```
igpetrov, line 21: Suggestion to add `equals` and `hashcode`. Just in case somebody wants to put it in map in the future.

igpetrov (approved):

igpetrov (approved):

core/src/main/java/io/camunda/connector/api/inbound/ProcessCorrelationPoint.java
```diff
@@ -14,18 +14,29 @@
...
  * limitations under the License.
  */
-package io.camunda.connector.runtime.util.inbound;
+package io.camunda.connector.api.inbound;
 
-import static org.junit.jupiter.api.Assertions.assertThrowsExactly;
+public abstract class ProcessCorrelationPoint implements Comparable<ProcessCorrelationPoint> {
 
-import org.junit.jupiter.api.Test;
+  protected final String bpmnProcessId;
```
berndruecker, line 21: I thought those properties would be moved to [StartEventInboundConnectorResult.java](https://github.com/camunda-community-hub/spring-zeebe/pull/317/files#diff-05567545735773dfadce805569809c59fd3309ca27ebf25928d87a9dca46f9a8)?

berndruecker (commented):
Hi @chillleader - the only comment i currently have is that I would expect those CorrelationPoint properties to be removed here.

At the same time I could imagine that [StartEventCorrelationPoint.java](https://github.com/camunda-community-hub/spring-zeebe/pull/317/files#diff-a7821271d25cd5e2eba09c656c7a379c0d1b47b7f1d2efb390419f09f50a79ce) and [MessageCorrelationPoint.java](https://github.com/camunda-community-hub/spring-zeebe/pull/317/files#diff-be56b0927b52ee3998cb00dbb8972e5af40850bf884273a18656a3992761b5d1) even be moved to the SDK itself - as they are "just" DTO objects now and then everything would be close together - WDYT?

core/src/main/java/io/camunda/connector/api/inbound/ProcessCorrelationPoint.java
chillleader: Did you mean [StartEventCorrelationPoint](https://github.com/camunda-community-hub/spring-zeebe/pull/317/files#diff-a7821271d25cd5e2eba09c656c7a379c0d1b47b7f1d2efb390419f09f50a79ce)?

If so, I think we need these properties in both `CorrelationPoint` types. As we are wiring the inbound connector lifecycle events (activation and deactivation) to new process definition version deployment, I think it is fair that `CorrelationPoint` always exists for a specific process definition only. Therefore we will likely reference these properties in the code that manages the lifecycle. WDYT?

Also, moved the implementation classes to the SDK 👍

berndruecker: I added a comment in Spring Zeebe as well - I am not a huge fan of "misusing" those properties from the correlation point to know which process definition defined the connector - maybe we should focus this on the information relevant to correlation only - and pass those information differently? Just an idea - feel free to ignore

berndruecker (dismissed):
Added some comments - but looks good to me overall!

core/src/main/java/io/camunda/connector/api/inbound/ProcessCorrelationPoint.java
chillleader: I agree, makes sense to have this information separated. Moved to `InboundConnectorProperties`

PR #362, opened 2023-02-01
Define Inbound connector interfaces

chillleader:
## Description

Defines basic interfaces and abstraction for inbound connectors:
- `InboundConnector` annotation (similar to `OutboundConnector`, to be used by runtime)
- `InboundConnectorExecutable` interface - the one that will be used by runtime and implemented by all inbound connectors)
- `SubscriptionInboundConnector` -  base class for subscription-like connectors. Defines specific methods.

## Related issues

inspired by #334


core/src/main/java/io/camunda/connector/api/inbound/InboundConnectorContext.java
```diff
@@ -45,6 +43,5 @@ public interface InboundConnectorContext {
    * @param variables - variables to be passed to the process
    * @return
    */
-  InboundConnectorResult correlate(
-      ProcessCorrelationPoint correlationPoint, Map<String, Object> variables);
+  InboundConnectorResult correlate(ProcessCorrelationPoint correlationPoint, Object variables);
```
berndruecker, line 46: Don't we use Map<String, Object> everywhere for variables? Not sure - just would make sense and I think we should make it consstent

chillleader: I was primarily looking at similar examples in the SDK, e.g. in the outbound stuff. Here's how it currently looks like:
- As an outbound connector developer, I declare custom DTO classes that stores my input/output variables.
- SDK provides input variables either already deserialized in an object ([`getVariablesAsType`](https://github.com/camunda/connector-sdk/blob/a50eaadd44c0b5aa58c1f0e9fa20dd42a78a80a1/core/src/main/java/io/camunda/connector/api/outbound/OutboundConnectorContext.java#L40)) or as a raw JSON string in case I need something specific.
- SDK accepts my output variables [as an Object too](https://github.com/camunda/connector-sdk/blob/a50eaadd44c0b5aa58c1f0e9fa20dd42a78a80a1/core/src/main/java/io/camunda/connector/api/outbound/OutboundConnectorFunction.java#L39).
- `replaceSecrets` method [consumes an object as well](https://github.com/camunda/connector-sdk/blob/a50eaadd44c0b5aa58c1f0e9fa20dd42a78a80a1/core/src/main/java/io/camunda/connector/api/outbound/OutboundConnectorContext.java#L47)

So it seems to me that I don't actually work with `Map` anywhere as an SDK user - therefore I think it makes sense to use an Object here.

berndruecker: OK - great! I might be influenced by my past experiences with the engine overall :-) Given that Object is more generic anyway - can't be too wrong I guess

PR #364, opened 2023-02-03
Inbound Connector Discovery

chillleader:
## Description

This PR implements inbound connector discovery, while also consolidating the discovery of outbound connectors.

Key changes:
- `ConnectorFactory` interface that encapsulates connector configuration discovery and connector instance creation.
  -  `InboundConnectorFactory` impl: creates instances on-demand. Unlike outbound connectors, which are essentially stateless functions, inbound connectors are **stateful** disposable objects.
  - `OutboundConnectorFactory` impl: maintains a pool of pre-instantiated outbound connector functions that are reused globally.
- Refactoring of `OutboundConnectorRegistrationHelper` class: it is broken down into `SPIConnectorDiscovery` and `EnvVarConnectorDiscovery` classes, and similar discovery methods for inbound connectors are implemented.
- Provided documentation for connector discovery

see also: https://github.com/camunda-community-hub/spring-zeebe/pull/331 for support of these changes in Spring Zeebe

related to https://github.com/camunda/team-connectors/issues/331


core/src/main/java/io/camunda/connector/api/inbound/InboundConnectorExecutable.java
```diff
@@ -39,5 +41,5 @@ void activate(InboundConnectorProperties properties, InboundConnectorContext con
    * Gentle shutdown hook for inbound connectors. Must release all resources used by the
    * subscription.
    */
-  void deactivate() throws Exception;
+  void deactivate(InboundConnectorProperties properties) throws Exception;
```
igpetrov, line 44: Not a big thing, but do we actually need to have the whole object `InboundConnectorProperties` to shut down?

runtime-util/README.md
```diff
@@ -38,6 +38,51 @@ public class Main {
 }
 ```
 
+## Connector discovery
+
+Implementations of tbe `ConnectorFactory` interface are responsible for Connector
```
igpetrov, line 43:
```suggestion
Implementations of the `ConnectorFactory` interface are responsible for Connector
```

runtime-util/src/main/java/io/camunda/connector/runtime/util/discovery/EnvVarsConnectorDiscovery.java
```diff
@@ -0,0 +1,160 @@
...
+public class EnvVarsConnectorDiscovery {
+
+  /** Pattern describing the outbound connector env configuration pattern. */
+  public static final Pattern OUTBOUND_CONNECTOR_FUNCTION_PATTERN =
+      Pattern.compile("^CONNECTOR_(.*)_FUNCTION$");
+
+  public static final Pattern INBOUND_CONNECTOR_EXECUTABLE_PATTERN =
+      Pattern.compile("^CONNECTOR_(.*)_EXECUTABLE$");
+
+  public static Map<String, String> hardwiredEnvironmentVariables;
```
igpetrov, line 43: I would rather not make it `public` but rather `private`.

runtime-util/src/main/java/io/camunda/connector/runtime/util/discovery/EnvVarsConnectorDiscovery.java
```diff
@@ -0,0 +1,160 @@
...
+
+  public static void addHardwiredEnvironmentVariable(String key, String value) {
+    if (hardwiredEnvironmentVariables == null) {
+      hardwiredEnvironmentVariables = new HashMap<>();
```
igpetrov, line 47: In addition, I would also make it as `ConcurrentHashMap` just in case, or make method synchronized.

runtime-util/src/test/java/io/camunda/connector/runtime/util/inbound/InboundConnectorDiscoveryTest.java
```diff
@@ -0,0 +1,153 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.camunda.connector.runtime.util.inbound;
+
+import static io.camunda.connector.runtime.util.TestUtil.*;
```
igpetrov, line 19: We don't use `*` imports.

igpetrov (dismissed):
LGTM, minor comments.

core/src/main/java/io/camunda/connector/api/inbound/InboundConnectorExecutable.java
chillleader: I added this is to potentially support 2 lifecycle models with this interface:
- Disposable isolated objects, i.e. normal subscription use case
- Singleton connector instance that manages all of the subscriptions, for special use cases like webhook connector

So it's like a tradeoff between flexibility of usage and unambiguity of the interface. I haven't completely made my mind about this yet, but I feel like flexibility might be more important.

igpetrov: Got it. Thank you!

Issue #393, opened 2023-03-03
Support cancellation callback in inbound Connectors API


chillleader:
__Is your feature request related to a problem? Please describe.__

Currently, inbound subscriptions are started by the runtime and stopped only by the runtime. But it's possible that subscription fails unexpectedly, e.g. due to network issues. For this case, we need a callback that will be able to signal the runtime that subscription has failed. Then, runtime can possible do some retries or report back to the user via Modeler.

__Describe the solution you'd like__

A new method in `InboundConnectorContext`: `InboundConnectorContext::cancel`.
It can accept either an error message or a raw exception, which then will be analyzed by the runtime.

__Describe alternatives you've considered__

* Do not do this (at least for the first version) - this may quickly become confusing for the user.
* Let the Connector code handle retries on its own - this way we will have many repetitions of retry code.

PR #411, opened 2023-03-30
Cancellation callback, activation condition and variable mappings

chillleader:
## Description

Introduced the following changes to Connector SDK:

- **Support cancellation callback for inbound connectors** (#393). This is needed to handle external errors that may happen with active inbound connectors, e.g. network errors that result in subscription failure.
- **Moved `activationCondition` and `variableMapping` properties handling to the SDK/Runtime level** (#412). Variable mapping field was replaced with 2 fields: `resultExpression` and `resultVariable`.
  - Now the Runtime supports a similar field set for inbound and outbound connectors for mapping variables (`resultVariable` and `resultExpression`, providing identical behavior for both types of connectors.
- Other general improvements:
  - Error handling was improved (now business errors like activation condition mismatch don't result in exception being thrown, plus replaced the generic `Object` type of `responseData` in `InboundConnectorResult`).
  - Minor adjustments to javadocs
  - Removed `SubscriptionInboundConnector` abstract class (likely not needed, doesn't add much value)

## Related issues

related to #393, #412


chillleader:
Need to rework a few things around `activationCondition` and `variableMapping`

igpetrov (approved):
LGTM. Since some classes changed their packages, do you expect this change to be with breaking changes?

If so, let's move it faster so that we have time to test runtime and connectors.

igpetrov:
also, @chillleader, shall we squash into 1 commit or do you want to keep those as 4?

chillleader:
@igpetrov there are a couple of changes required in spring-zeebe after this, I'm planning to open a PR in spring-zeebe today or tomorrow supporting these changes.

igpetrov:
@chillleader perfect, thanks for explanation. We just found an issue with Spring Zeebe migration in alpha3 build - breaking changes from [PR](https://github.com/camunda-community-hub/spring-zeebe/pull/372) but @markfarkas-camunda is already on it.

PR #430, opened 2023-04-13
fix(inbound): handle legacy variable mapping for webhook

### LINKED ISSUE #429: Legacy variable mapping doesn't work for webhook Connector
__Describe the Bug__

After the 0.8.0 release, the legacy variable mapping property (`inbound.variableMapping`) stopped working for webhook connector.

The root cause is that we introduced new common inbound connector properties for the runtime that replace the old variable mapping: `resultExpression` and `resultVariable`. However, webhook connector still uses the legacy variable, and we should handle this as an exceptional case.

__Steps to Reproduce__

1. Model a process with webhook connector. Specify variable mapping.
1. Deploy the process, run the connector runtime with SDK 0.8.0
2. Send a webhook request with an appropriate body that suits the variable mapping
3. Make sure that the process is started, and no variables are in the process context, i.e. variable mapping was evaluated successfully but didn't take effect.

__Expected Behavior__

If legacy variable mapping is used, it should be processed by the webhook connector, and the runtime components should not process result expressions in that case.

chillleader:
## Description

* Adds handling of a special case when legacy variable mapping is used in webhook connector (instead of the new common `resultVariable` and `resultExpression` properties common for all inbound connectors)
* Also fixes some inconsistency between submodule versions (all should be 0.8.0-SNAPSHOT at this point)

## Related issues

closes #429


igpetrov (approved):

igpetrov:
@chillleader, so, seems @bastiankoerber was using old element template?

PR #431, opened 2023-04-13
feat(inbound): support nested inbound properties

chillleader:
## Description

Adds support for nested inbound connector properties.

In outbound connector inputs we often use variables like `authentication.uri="http://localhost"` which get transformed into structures like `authentication: { uri: "http://localhost" }` by the Zeebe engine before entering Connectors.

This allows us to deserialize connector inputs as complex objects, containing other nested objects.

For inbound connector properties, this is currently not supported - this PR adds a support for similar mechanism so that we can deserialize inbound connector properties in a similar fashion with nested objects.

## Related issues

related to https://github.com/camunda/team-connectors/issues/351


igpetrov:
Just to confirm there's no mistake. The branch is `rabbitmq-updates`.

PR #435, opened 2023-04-18
feat(inbound): wrap FEEL errors in ConnectorInputException

chillleader:
## Description

Wraps FEEL engine errors that happen during inbound event correlation in `ConnectorInputException` so that such errors can be caught and handled differently in the connector implementation. (e.g. avoid re-queueing messages that can not be deserialized.


core/src/main/java/io/camunda/connector/test/inbound/InboundConnectorPropertiesBuilder.java
```diff
@@ -111,6 +111,14 @@ public InboundConnectorPropertiesBuilder property(String key, String value) {
     return this;
   }
 
+  public InboundConnectorPropertiesBuilder properties(Map<String, String> newProperties) {
```
igpetrov, line 114: Consider adding a null-safety for `newProperties`

igpetrov (approved):
LGTM

Issue #462, opened 2023-05-16
Inbound connector health indicator support


sbuettner:
The `InboundConnectorContext` should provide basic methods to allow inbound connectors to report their health status including additional data that allows users to investigate failing inbound connectors similar to Spring Boot´s [Health](https://docs.spring.io/spring-boot/docs/current/api/org/springframework/boot/actuate/health/Health.html) implementation.

- [x] Provide based methods to report health status and data on the context level
- [ ] Use this data in the runtime and expose it via the `InboundConnectorRestController`.

sbuettner:
Relates to: https://github.com/camunda/team-connectors/issues/371

PR #463, opened 2023-05-16
Provide a facility for inbound connectors to report their health status

sbuettner:
## Description

This PR add a facility for inbound connectors to report their health status via the context.

## Related issues

relates to #462


core/src/main/java/io/camunda/connector/api/inbound/InboundConnectorContext.java
```diff
@@ -85,4 +85,17 @@ public interface InboundConnectorContext {
...
+   * data can be used to report data on connectiveness, liveliness and whether the Connector is
+   * running successfully.
+   *
+   * <p>This method can be called as often as needed and the internal state of the inbound Connector
+   * implementation requires it.
+   *
+   * @param @{@link Health} health of the inbound connector including optional details about the
+   *     status.
+   */
+  void report(Health health);
```
chillleader, line 100: nit: maybe we can call it `reportStatus` to be more explicit, otherwise someone might confuse this method's purpose with `correlate` method (whose name is also vague and perhaps not very intuitive).

markfarkas-camunda (approved):
Looks good to me! 👍

chillleader (approved):

core/src/main/java/io/camunda/connector/api/inbound/InboundConnectorContext.java
sbuettner: @chillleader What about `reportHealth`?

PR #466, opened 2023-05-16
feat: add element id to InboundConnectorProperties

chillleader:
## Description

Adds `elementId` to inbound connector properties to enable filtering by specific elements in the web modeler.

## Related issues

related to https://github.com/camunda/connectors-bundle/issues/591


sbuettner (approved):

core/src/main/java/io/camunda/connector/test/inbound/InboundConnectorContextBuilder.java
```diff
@@ -166,7 +166,8 @@ private InboundConnectorProperties getGenericProperties() {
         new HashMap<>(),
         UUID.randomUUID().toString(),
         0,
-        0);
+        0,
+        "myMsgCatchEvent");
```
markfarkas-camunda, line 170: Just a thought: since we just use a default String here, can't we use "defaultElementId" instead of "myMsgCatchEvent"? That would give a hint about what this is exactly.
Also wondering: do we ever want to use this method? Does it make sense to have an InboundConnectorProperties instance with default attributes?

core/src/main/java/io/camunda/connector/test/inbound/InboundConnectorPropertiesBuilder.java
```diff
@@ -35,6 +35,7 @@ public class InboundConnectorPropertiesBuilder {
   private String bpmnProcessId = "test-process";
   private int version = 1;
   private long processDefinitionKey = 1;
+  private String elementId = "test-element";
```
markfarkas-camunda, line 38: Why do we need the default value here?

markfarkas-camunda (approved):
Looks good to me, added two comments just for clarification.

igpetrov (approved):

core/src/main/java/io/camunda/connector/test/inbound/InboundConnectorPropertiesBuilder.java
chillleader: In some test cases, we might not care about element id. Then this shortcut might be useful to avoid extra meaningless lines of code in tests.

PR #471, opened 2023-05-17
feat(webhook): extend support for webhooks

igpetrov:
## Description

This code adjusts SDK in order to rework HTTP Webhook support.

Part 1/2, part 2: https://github.com/camunda/connectors-bundle/pull/603


core/src/main/java/io/camunda/connector/api/inbound/webhook/WebhookResponsePayload.java
```diff
@@ -0,0 +1,51 @@
...
+   */
+  Map<String, Object> params();
+
+  /**
+   * @return additional connector data that can be computed during webhook {@link
+   *     WebhookConnectorExecutable#triggerWebhook(InboundConnectorContext, WebhookRequestPayload)}
+   *     execution, and the implementer wants to preserve that data and use it during the {@link
+   *     io.camunda.connector.api.inbound.InboundConnectorContext#correlate(Object)} phase.
+   */
+  Map<String, Object> connectorData();
```
sbuettner, line 50: From an implementors endpoint its not clear whether the `connectorData` will be correlated by the runtime as `triggerWebhook` itself has access to the `InboundConnectorContext`. I think we could remove the `InboundConnectorContext` from the `triggerWebhook` to make it more clear. WDYT?

igpetrov: Or did you mean, we initialize it as a class member during `activate`?

igpetrov: That's actually a good idea, let me check

igpetrov: That worked out nicely! Thanks for suggestion @sbuettner.

core/src/main/java/io/camunda/connector/api/inbound/webhook/WebhookResponsePayload.java
chillleader, line 1: If I understand correctly, objects of this class are used as payload data for event correlation, which in turn happens in the runtime and not in the webhook connector implementation.

If this is true, the name `WebhookResponsePayload` seems to be a bit confusing: since we call it "Response payload", I would first assume that this is the value that will be returned by the webhook endpoint, when in fact it is a value that will be passed to the process instance.

Should we call it something like `WebhookEventPayload` / `WebhookCorrelationPayload` / `WebhookProcessingResult` / etc.. ? (My first random naming ideas are usually not the best, but maybe you can come up with something better 😄 )

core/src/main/java/io/camunda/connector/api/inbound/webhook/WebhookRequestPayload.java
```diff
@@ -0,0 +1,46 @@
...
+
+  /**
+   * @return HTTP Request method as {@link String}
+   */
+  String method();
+
+  /**
+   * @return HTTP Request headers as {@link Map}
+   */
+  Map<String, Object> headers();
```
chillleader, line 32: Why `Map<String, Object`? Can headers be something other than String?

igpetrov: Theoretically, it can be non-UTF-8 encoded "string" that in java `String` wouldn't be the same. I guess, what you mean, does make sense to keep it `<String, String>` from practical point of view.

core/src/main/java/io/camunda/connector/api/inbound/webhook/WebhookResponsePayload.java
igpetrov: Done. Also renamed the `RequestPayload` object.

core/src/main/java/io/camunda/connector/api/inbound/webhook/WebhookRequestPayload.java
igpetrov: Done, also applied the same to `params`.

chillleader (approved):
Thank you for the changes, lgtm!

core/src/main/java/io/camunda/connector/api/inbound/webhook/WebhookResponsePayload.java
igpetrov: @sbuettner kind reminder

sbuettner: > Or did you mean, we initialize it as a class member during activate?

Exactly.

sbuettner (approved):
Thanks for taking care @igpetrov. LGTM!

PR #474, opened 2023-05-22
chore(naming): Rename GCP secret provider

### LINKED ISSUE #473: Remove custom version from connector-gcp-security-manager
__What should we do?__

All other artifacts use the version of the parent. The same should apply for the `connector-gcp-security-manager` artifact. 

__Why should we do it?__

Consistent versioning of SDK artifacts.

sbuettner:
closes #473


igpetrov (approved):
I am wondering whether users can use this module separately? If so, we need to mention the fact that we renamed module into the release migration guide.

chillleader:
`secret-provider/gcp-security-manager` actually makes sense to me, since it's a secret provider implementation that sources secret data from GCP Security Manager. But I don't have a strong preference so feel free to rename it if you feel like it.

chillleader (approved):

sbuettner:
Good point @igpetrov but I dont think its part of the product offering right now.

Issue #247, opened 2022-11-10
Allow access to headers


jwulf:
__Is your feature request related to a problem? Please describe.__

Not sure if I'm missing something, but it looks like there is no way to get access to headers in a connector?

__Describe the solution you'd like__

Maybe a `getHeaders` method, similar to `getVariablesAsType`, to allow access to task headers.

__Describe alternatives you've considered__

Don't use headers, only variables.

nikku:
Could you elaborate where you'd need headers? 

Headers have the disadvantage to be static only and in most usage scenarios that is not desired. Inputs on the other hand can be static values `hello` or expressions `= foobar + "!"`, improving the expressiveness of templates.

jwulf:
Do the variables from the connector template get merged into the payload past the connector? 

Headers are scoped to the task, so you don't have to worry about namespacing and collisions/overwriting.

nikku:
> Headers are scoped to the task, so you don't have to worry about namespacing and collisions/overwriting.

That is an upside. The idea is that variable merging is prevented through namespacing, however we've already received indications that our existing implementation is prone to unintended "clashes".

MaxTru:
Given that headers have the clear scoping this is a valid use-case. Moving it to backlog.

igpetrov:
Taking a look

igpetrov:
Added headers, and JSON representation of complete activated job for those who need more flexibility.

https://github.com/camunda/connector-sdk/pull/446

PR #501, opened 2023-06-29
First draft for simplified object mapping

sbuettner:
First draft for simplified object mapping and secret handling.

## Description

The idea is to unify json mapping, secret replacement and validation.

## Related issues

relates to camunda/team-connectors/issues/392


core/src/main/java/io/camunda/connector/impl/context/AbstractConnectorContext.java
```diff
@@ -55,13 +58,6 @@ public void validate(Object input) {
...
    */
   protected ValidationProvider getValidationProvider() {
-    return ServiceLoader.load(ValidationProvider.class)
-        .findFirst()
-        .orElseThrow(
-            () ->
-                new IllegalStateException(
-                    "Please bind an implementation to "
-                        + ValidationProvider.class.getName()
-                        + " via SPI"));
```
chillleader, line 65: Why not keep the default SPI discovery for validation provider?
I assume we want to use the Spring context, but given that successors can access the field directly, do we need this method at all?

core/src/main/java/io/camunda/connector/api/outbound/OutboundConnectorContext.java
```diff
@@ -29,44 +27,7 @@ public interface OutboundConnectorContext {
...
-   * @return the mapped object containing the variable values
-   */
-  <T> T getVariablesAsType(Class<T> cls);
-
-  /**
-   * Custom headers found under zeebe:taskHeaders.
-   *
-   * @return custom job headers.
-   */
-  Map<String, String> getCustomHeaders();
```
chillleader, line 49: Was it deleted on purpose? Looks like this is not related to variable binding.

core/src/main/java/io/camunda/connector/api/outbound/OutboundConnectorContext.java
```diff
@@ -29,44 +27,7 @@ public interface OutboundConnectorContext {
...
-   */
-  void replaceSecrets(Object input);
+  <T> T bindVariables(Class<T> cls);
 
-  /**
-   * Validates the input object
-   *
-   * @param input - the object to validate
-   */
   void validate(Object input);
```
chillleader, line 32: This method is also going to be deleted, right?

markfarkas-camunda (changes requested):
For me tests in OutboundConnectorContextTest.java are failing. Could you please check those?

core/src/main/java/io/camunda/connector/api/annotation/Secret.java
```diff
@@ -32,6 +32,7 @@
  */
 @Target(ElementType.FIELD)
 @Retention(RetentionPolicy.RUNTIME)
+@Deprecated(forRemoval = true)
```
igpetrov, line 35: Suggestion: let's also add a javadoc for deprecation note to avoid confusion from external contributors like "ok, I see it's deprecated, what should I use instead?".

chillleader (commented):
I like the direction 👍 `bindVariables` is much simpler and probably more intuitive. But I think we should also keep some lower-level data access methods, e.g. `OutboundConnectorContext::asJson`.

Are you planning to do the same for inbound connectors by introducing the `bindProperties` method in `InboundConnectorContext`?

core/src/main/java/io/camunda/connector/api/outbound/OutboundConnectorContext.java
igpetrov, line 49: This was a request from one of the customers at some point. They added custom task headers and wanted to get those from the context.

core/src/main/java/io/camunda/connector/api/secret/SecretContainerHandler.java
```diff
@@ -20,6 +20,7 @@
  * Handler for secret containers. This allows to define a strategy for traversing the elements in a
  * secret container, e.g. using an iterator or for loop to handle all elements.
  */
+@Deprecated(forRemoval = true)
```
igpetrov, line 23: Same here. Let's leave a deprecation notice why it's deprecated and what to do instead.

core/src/main/java/io/camunda/connector/api/outbound/OutboundConnectorContext.java
sbuettner: I just removed things that werent used anywhere.

sbuettner: Good catch. This was unclear as no connector used it on our side.

sbuettner: But as @igpetrov said, apparently this is requests. Will bring it back :)

igpetrov: Yep, see this one https://github.com/camunda/connector-sdk/issues/247

core/src/main/java/io/camunda/connector/api/outbound/OutboundConnectorContext.java
```diff
@@ -24,49 +24,19 @@
...
-  void validate(Object input);
+  <T> T bindVariables(Class<T> cls);
 
-  /**
-   * A method intended for users who need to activated job in its complete JSON representation.
-   * Should be used as a final option when other APIs cannot supply a necessary data.
-   *
-   * @return JSON representation of an activated job.
-   */
-  String asJson();
```
chillleader, line 71: I'd be a bit hesitant to delete this. There's a chance someone is using this method 🙂

core/src/main/java/io/camunda/connector/impl/context/AbstractConnectorContext.java
```diff
@@ -19,18 +19,21 @@
...
-  protected AbstractConnectorContext(final SecretProvider secretProvider) {
+  protected final ValidationProvider validationProvider;
+
+  protected AbstractConnectorContext(
+      final SecretProvider secretProvider, final ValidationProvider validationProvider) {
     if (secretProvider == null) {
       throw new RuntimeException("Secret provider required in Connector context but was null");
     }
     this.secretProvider = secretProvider;
+    this.validationProvider = validationProvider;
```
chillleader, line 36: nit: for consistency, we can also check if `validationProvider` is null, similarly to `secretProvider`

sbuettner: @chillleader The idea was to make it optional. WDYT?

sbuettner:
@igpetrov @markfarkas-camunda @chillleader Please take another look.

core/pom.xml
```diff
@@ -16,6 +16,12 @@
 
   <dependencies>
 
+    <dependency>
+      <groupId>com.fasterxml.jackson.core</groupId>
+      <artifactId>jackson-databind</artifactId>
+      <scope>test</scope>
+    </dependency>
```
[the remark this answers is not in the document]
sbuettner: @chillleader Thinking about this one: Should we maybe keep it as a dependency so users can use jackson annotations in their request model classes? This would indicate that its always the default json implementation. WDYT?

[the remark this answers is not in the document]
chillleader: Makes sense 👍 I'm thinking that we should maybe add it with the `provided` scope then to avoid version conflicts between connectors and the runtime, WDYT?

core/src/main/java/io/camunda/connector/impl/context/AbstractConnectorContext.java
```diff
@@ -55,13 +58,6 @@ public void validate(Object input) {
    * @return the desired validation provider implementation
    */
   protected ValidationProvider getValidationProvider() {
```
chillleader, line 60: The javadoc for this method needs to be updated:

> By default, SPI is being used and **should be implemented by each implementation**.

This doesn't make sense any more since SPI discovery has been moved to the runtime

core/src/main/java/io/camunda/connector/impl/context/AbstractConnectorContext.java
chillleader: But in case it's not provided, wouldn't it lead to an NPE in `validate` method of this class (which may be relied upon by the ConnectorContext)

sbuettner: @chillleader I would be fine to make it none-optional then. WDYT?

chillleader (approved):
Awesome. To summarize the remaining items: we need to update javadocs, and the `validate` method is subject to removal. I'm ok with completing those as follow-ups (e.g. in my upcoming PR about inbound properties).

sbuettner:
Thank you @chillleader. Any concerns, ideas or approvals from @markfarkas-camunda or @igpetrov ?

igpetrov (approved):

PR #504, opened 2023-07-03
feat(inbound): simplified input mappings for inbound connectors

chillleader:
## Description

Redefined inbound connectors API to match with the new simplified input mappings implemented for outbound connectors.

Support of these changes in the connectors-bundle: https://github.com/camunda/connectors-bundle/pull/830

## Related issues

related to https://github.com/camunda/team-connectors/issues/392
related to https://github.com/camunda/connectors-bundle/issues/821


core/src/main/java/io/camunda/connector/api/inbound/InboundConnectorContext.java
```diff
@@ -68,23 +54,43 @@ public interface InboundConnectorContext {
...
-   * type. Deserialization logic is runtime-specific. If you need a lower-level access to properties
-   * (e.g. for custom deserialization), use {@link #getProperties()}
+   * Provides an object that references the process definition that the inbound Connector is
+   * configured for. The object can be used to access the process definition metadata.
    *
-   * @return - Connector-specific properties deserialized to a provided type
+   * @return definition of the inbound Connector
    */
-  <T> T getPropertiesAsType(Class<T> cls);
+  InboundConnectorDefinition getDefinition();
```
sbuettner, line 93: Like that we hide the details here. We could offer something similar for a Job (instead of providing everything as json) for outbound connectors.

core/src/main/java/io/camunda/connector/api/inbound/InboundConnectorDefinition.java
```diff
@@ -0,0 +1,40 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.camunda.connector.api.inbound;
+
+import io.camunda.connector.impl.inbound.ProcessCorrelationPoint;
+
+public interface InboundConnectorDefinition {
```
sbuettner, line 21: What about adding the `type` that we hide in the properties via `INBOUND_TYPE_KEYWORD` and make it available via this interface.

chillleader: Do you have a use-case in mind for exposing the `type` here? I couldn't come up with a valid one, because connector should already know its own type from its static configuration. And inside the connector runtime, we can just extract it from properties.

sbuettner: But wouldnt it be also nicer for the runtime to make use of that information when trying to resolve the type? As an example the `ActivatedJob` interface also provides access to the `type` via the `String getType();` method.

chillleader: Yeah, I agree. Still it feels somewhat unnatural, as the idea of this change was to separate stuff that is actually properties from things that don't come from properties that were previously part of the properties object nonetheless. 

I'll check how it might work out in the runtime and get back to this.

chillleader: Done

core/src/main/java/io/camunda/connector/api/outbound/OutboundConnectorContext.java
```diff
@@ -32,11 +32,39 @@ public interface OutboundConnectorContext {
...
+   * @return deserialized and validated variables with secrets replaced
+   */
   <T> T bindVariables(Class<T> cls);
 
+  /**
+   * Deprecated: use {@link #bindVariables(Class)} instead, where validation is performed
+   * automatically.
+   */
+  @Deprecated(forRemoval = true)
   void validate(Object input);
```
sbuettner, line 69: Maybe we can remove it already?

test/src/main/java/io/camunda/connector/test/inbound/InboundConnectorContextBuilder.java
```diff
@@ -89,61 +107,80 @@ public InboundConnectorContextBuilder property(String key, String value) {
...
+    }
+    properties.put(RESULT_VARIABLE_KEYWORD, variable);
+    return this;
+  }
+
+  public InboundConnectorContextBuilder resultExpression(String expression) {
+    if (properties.containsKey(RESULT_EXPRESSION_KEYWORD)) {
+      throw new IllegalArgumentException("Result variable already set");
+    }
+    properties.put(RESULT_EXPRESSION_KEYWORD, expression);
```
sbuettner, line 171: What about we model it somehow using a proper model (imlpementation of `InboundConnectorDefinition`)  and remove the "SDK" properties from the "userspace" properties? Otherwise customers could "bind" these properties.

test/src/main/java/io/camunda/connector/test/inbound/InboundConnectorContextBuilder.java
```diff
@@ -89,61 +107,80 @@ public InboundConnectorContextBuilder property(String key, String value) {
    * @param properties - new properties
    * @return builder for fluent API
    */
-  public InboundConnectorContextBuilder properties(InboundConnectorProperties properties) {
-    assertNoProperties();
+  public InboundConnectorContextBuilder properties(Map<String, Object> properties) {
+    if (this.properties != null) {
+      throw new IllegalStateException("properties already set");
```
igpetrov, line 112: nit:
```suggestion
      throw new IllegalStateException("Properties already set");
```

igpetrov: I am wondering, if property already set, do we need to interrupt the flow? Maybe we can continue the flow, or at least check if old and new properties are the same, and if so, continue? Just thinking out loud.

igpetrov (approved):
LGTM

test/src/main/java/io/camunda/connector/test/inbound/InboundConnectorContextBuilder.java
chillleader: Good idea

test/src/main/java/io/camunda/connector/test/inbound/InboundConnectorContextBuilder.java
chillleader: Added an equality check, in general I think it's fine to enforce that user only defines mock properties once.

core/src/main/java/io/camunda/connector/impl/secret/SecretHandler.java
```diff
@@ -214,6 +215,13 @@ protected static boolean isPrimitive(Class<?> clazz) {
...
   public String replaceSecrets(String input) {
-    return SecretUtil.replaceSecrets(input, secretProvider::getSecret);
+    Function<String, String> secretReplacer =
+        name ->
+            Optional.ofNullable(secretProvider.getSecret(name))
+                .orElseThrow(
+                    () ->
+                        new IllegalArgumentException(
+                            String.format("Secret with name '%s' is not available", name)));
+    return SecretUtil.replaceSecrets(input, secretReplacer);
```
sbuettner, line 225: Can we extract and re-use an instance instead of recreating this function for every method call?

core/src/main/java/io/camunda/connector/impl/secret/SecretHandler.java
```diff
@@ -52,8 +52,17 @@ public class SecretHandler implements SecretElementHandler, SecretContainerHandl
...
+  protected Function<String, String> secretReplacer;
+
   public SecretHandler(final SecretProvider secretProvider) {
     this.secretProvider = secretProvider;
+    secretReplacer =
+        name ->
+            Optional.ofNullable(secretProvider.getSecret(name))
+                .orElseThrow(
+                    () ->
+                        new IllegalArgumentException(
```
sbuettner, line 64: One last nit, could you throw a custom exception here? This way we could catch it and know we dont expose any information and map it properly. Something like like `SecretException` or `SecretMappingException`?

sbuettner (approved):

sbuettner:
Great work. LGTM!

PR #507, opened 2023-07-05
Introduce response body expression

igpetrov:
Add `responseBodyExpression`

Related to https://github.com/camunda/connectors-bundle/pull/832


core/src/main/java/io/camunda/connector/impl/Constants.java
chillleader, line 1: What do you think about implementing the response body expression entirely in the runtime/webhook impl? It looks like this functionality won't be shared with other inbound connector types. Maybe we have already discussed this - then sorry if I missed 🙂

igpetrov: Yeah, that's the approach. The whole implementation is in the runtime / webhook: https://github.com/camunda/connectors-bundle/pull/832.

Here I just define a reserved key word essentially. Or maybe I didn't get the concern?

chillleader: I use this Constants class in my PR to distinguish runtime-level properties from connector-level properties and treat them differently (I added a collection [here](https://github.com/camunda/connector-sdk/blob/4833914b7fe18a25dc53007859ffa9859aa7a888/core/src/main/java/io/camunda/connector/impl/Constants.java#L79) and I will consume it in the runtime). The idea is that runtime-level properties are inaccessible to connectors. So I would vote for defining this keyword in the runtime too.

igpetrov: I just don't get how is it different to, say `correlationKeyExpression` or `errorExpression` and others? 🤔 
Aren't they also runtime-only and connector doesn't need to know about those?

---

Is your proposal to move the constant entirely to runtime? If so, I am okay with that, but it sounds to be sort of detached, like we have a special key word not defined with the rest of other keywords. Besides, won't we still need to keep it in [reserved key words](https://github.com/camunda/connector-sdk/blob/4833914b7fe18a25dc53007859ffa9859aa7a888/core/src/main/java/io/camunda/connector/impl/Constants.java#L79)?

chillleader: Right, I agree - then I'm good with adding it to `Constants` and keeping it away from the reserved keywords set.

chillleader (approved):

igpetrov:
@chillleader okay, maybe let's discuss it at todays daily sync?

sbuettner (approved):

sbuettner (approved):

sbuettner (approved):

PR #527, opened 2023-07-18
Add webhook result context & enhance FEEL deserializer

chillleader:
## Description

- Adds webhook result context to the SDK (so that it's accessible also in the webhook impl, needed for https://github.com/camunda/connectors-bundle/pull/816)
- Enhances FEEL deserializer so that it's capable of handling fields where FEEL is optional


core/src/main/java/io/camunda/connector/impl/feel/AbstractFeelDeserializer.java
```diff
@@ -26,10 +26,12 @@
 public abstract class AbstractFeelDeserializer<T> extends StdDeserializer<T>
     implements ContextualDeserializer {
   protected FeelEngineWrapper feelEngineWrapper;
+  protected boolean strict;
 
-  protected AbstractFeelDeserializer(FeelEngineWrapper feelEngineWrapper) {
+  protected AbstractFeelDeserializer(FeelEngineWrapper feelEngineWrapper, boolean strict) {
```
igpetrov, line 31: FYI: strict is being removed [here](https://github.com/camunda/connector-sdk/pull/507).

Not merging it, because it is blocked by [this PR](https://github.com/camunda/connectors-bundle/pull/832), otherwise it will break bundle.

chillleader: It's a different `strict` - I will rename it to avoid confusion

igpetrov: Ahhh good. No, you can keep it. The other one will be removed anyway

igpetrov (approved):

PR #531, opened 2023-07-19
feat(webhook): add body as map

igpetrov:
feat(webhook): add body as map


core/src/main/java/io/camunda/connector/api/inbound/webhook/WebhookProcessingPayload.java
```diff
@@ -31,6 +31,11 @@ public interface WebhookProcessingPayload {
    */
   String method();
 
+  /**
+   * @return HTTP Request body as {@link Map}
+   */
+  Map<String, Object> body();
```
sbuettner, line 37: What if the "body" is just a string?

sbuettner: @igpetrov What about a json value like a plain json string `"my body"` and/or a list like `[1,2,3]`. Shouldnt we also support that? Therefore maybe `Object` would be fine.

igpetrov: Removed in new version

igpetrov:
@sbuettner please take a look at another round and its [related bundle implementation](https://github.com/camunda/connectors-bundle/pull/925).

Thank you.

core/src/main/java/io/camunda/connector/api/inbound/webhook/WebhookProcessingResult.java
```diff
@@ -27,21 +28,9 @@
 public interface WebhookProcessingResult {
```
sbuettner, line 28: Can we rename it to `WebhookResult`?

core/src/main/java/io/camunda/connector/api/inbound/webhook/WebhookProcessingResult.java
```diff
@@ -52,4 +41,8 @@ default Map<String, String> params() {
   default Map<String, Object> connectorData() {
     return Collections.emptyMap();
   }
+
+  default Function<WebhookProcessResultContext, Object> responseBodyExpression() {
```
sbuettner, line 45: Can we rename it to `WebhookResultContext`?

sbuettner (commented):
Looks good to me just a naming nit for consistency.

core/src/main/java/io/camunda/connector/api/inbound/webhook/WebhookProcessingResult.java
igpetrov: Can do that.

core/src/main/java/io/camunda/connector/api/inbound/webhook/WebhookProcessingResult.java
igpetrov: I think, that would be not nice, because we have `WebhookProcessResultContext` and `WebhookTriggerResultContext`. WDYT @sbuettner ?

chillleader (approved):
LGTM

core/src/main/java/io/camunda/connector/api/inbound/webhook/WebhookProcessingResult.java
igpetrov: Consider the following snippet:

```
      var webhookResult =
          ((WebhookConnectorExecutable) connector.executable()).triggerWebhook(payload);
      var ctxData = toWebhookTriggerResultContext(webhookResult);
      InboundConnectorResult<?> result = connector.context().correlate(ctxData);
      var processVariablesContext = toWebhookProcessResultContext(webhookResult, result);
      var httpResponseData = webhookResult.responseBodyExpression().apply(processVariablesContext);
      connectorResponse = ResponseEntity.ok(httpResponseData);
```

core/src/main/java/io/camunda/connector/api/inbound/webhook/WebhookProcessingResult.java
igpetrov: Done

core/src/main/java/io/camunda/connector/api/inbound/webhook/WebhookProcessingResult.java
igpetrov: - `WebhookTriggerResultContext` result returned from webhook `triggerWebhook` method. The data will be supplied to Zeebe.
- `WebhookProcessResultContext` result returned from runtime - contains webhook process variables + correlation result. It will be post-processed by a response expression function to refine the final result.

> What about `WebhookResponseContext`?

The webhook response is close to `WebhookTriggerResultContext` - this what is returned from actual webhook execution, so I am not in favor of this name.

> As this is what the response is based on?

Yes, but this is not really a webhook response. It's runtime response. It contains both webhook and zeebe correlation result. It can potentially contain other variables if we wish to implement those at some point.

sbuettner: My point was to rename `WebhookProcessResultContext` to `WebhookResponseContext` as this is the context where the response is evaluated.

sbuettner: As `WebhookProcessResultContext` has nothing to do with a process. Its the combination of correlation, request and connectorData.

igpetrov: For me, both names are okay-ish. Naming is hard, you know :)

Updated.

sbuettner: It is :)

sbuettner (approved):

PR #534, opened 2023-07-21
feat: add plain JSON support in FEEL deserializer

chillleader:
## Description

Introducing the full JSON support for the FEEL deserializer. With these changes it is possible for the users to provide FEEL or JSON in the fields annotated with `@FEEL`.


core/src/main/java/io/camunda/connector/impl/feel/AbstractFeelDeserializer.java
```diff
@@ -37,22 +39,23 @@ protected AbstractFeelDeserializer(FeelEngineWrapper feelEngineWrapper, boolean
...
-        return doDeserialize(value);
-      }
+    if (isFeelExpression(node.textValue()) || relaxed) {
+      return doDeserialize(node, mapper);
     }
     throw new IOException(
-        "Invalid input: expected a FEEL expression, but got '" + node + "' instead.");
+        "Invalid input: expected a FEEL expression, but got '"
+            + parser.getCodec().readTree(parser)
+            + "' instead.");
```
sbuettner, line 53: Could this leak secrets?

chillleader: Totally. Replaced it with a property name.

sbuettner (approved):

PR #539, opened 2023-07-24
Support legacy string list format.

sbuettner:
## Description

Support legacy list format from strings like: a,b,c | 1,2,3 in @FEEL annotated properties.
 
## Related issues

Required for [camunda/connectors-bundle](https://github.com/camunda/connectors-bundle/pull/926)


core/src/main/java/io/camunda/connector/impl/feel/FeelDeserializer.java
```diff
@@ -53,17 +57,49 @@ protected Object doDeserialize(JsonNode node, ObjectMapper mapper)
...
-        // string)
-        return mapper.readValue(node.textValue(), outputType);
-      } catch (IOException e) {
-        // ignore, this is just a string, we will take care of it below
+      var textValue = node.textValue();
+      if (outputType.isCollectionLikeType()
+          && outputType.hasContentType()
+          && !textValue.trim().startsWith("[")) {
+        // Support legacy list like formats like: a,b,c | 1,2,3
+        if (outputType.getContentType().hasRawClass(Long.class)) {
```
chillleader, line 65: nit: do we also need e.g. integers here? I assume it's ok to just support longs, since this is a legacy format.

chillleader (approved):
LGTM

core/src/main/java/io/camunda/connector/impl/feel/FeelDeserializer.java
sbuettner: Good point, will add support.

sbuettner: Added support

chillleader (approved):

PR #540, opened 2023-07-24
feat: test context API improvements

chillleader:
## Description

Improvements in `OutboundConnectorContextBuilder` and `InboundConnectorContextBuilder` APIs. Now 3 types of property/variable assignment are supported:
  - `property(string, string)`
  - `properties(object)` (with special handling for immutable maps like `Map.of()`)
  - `properties(jsonString)`

Other changes:
- `JacksonModuleFeelFunction` has been moved to the core SDK to be available in the test contexts.
- `@FEEL` annotation moved to `io.camunda.connector.api.annotation` (annotations should be parts of API definition).
- Improved test coverage for test contexts.

## Related issues

closes #


test/src/main/java/io/camunda/connector/test/inbound/InboundConnectorContextBuilder.java
```diff
@@ -42,7 +48,14 @@ public class InboundConnectorContextBuilder {
   protected InboundConnectorResult<?> result = new MessageCorrelationResult("mockMsg", 0);
   protected ValidationProvider validationProvider;
 
-  protected ObjectMapper objectMapper = new ObjectMapper();
+  protected ObjectMapper objectMapper =
```
sbuettner, line 51: Can use a shared default for all object mapper usages? Maybe from the core?

test/src/main/java/io/camunda/connector/test/outbound/OutboundConnectorContextBuilder.java
```diff
@@ -34,12 +40,18 @@ public class OutboundConnectorContextBuilder {
...
-  private final ObjectMapper objectMapper =
-      new ObjectMapper().configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
+  private ObjectMapper objectMapper =
+      new ObjectMapper()
+          .registerModule(new JacksonModuleFeelFunction())
+          .registerModule(new Jdk8Module())
+          .registerModule(new JavaTimeModule())
+          // deserialize unknown types as empty objects
+          .disable(SerializationFeature.FAIL_ON_EMPTY_BEANS)
+          .disable(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES);
```
sbuettner, line 54: Same as above. Lets use a shared one from core here.

sbuettner (approved):

test/src/main/java/io/camunda/connector/test/inbound/InboundConnectorContextBuilder.java
chillleader: Let's do this when we move the SDK to the bundle. I don't want to include Jackson dependencies to runtime core (currently they are `provided` but defining a mapper there will require to include them in a normal way, making the connector jars too fat and causing dependency conflicts potentially).

I would like to change the structure a bit and have a separate artifact with API only and another one with Jackson.

